PR #4, opened 2021-10-08
Protocol of uploaded file based on X-Forwarded-Proto

Daedren:
When a file's uploaded, the file URL returned comes with either "http://" or "https://" depending if TLS is being used.
If you're using a reverse proxy, it'll return http on the URL even if you're actually connecting to the proxy through https (and then http to copyparty)

X-Forwarded-Proto or X-Forwarded-Ssl seem like the most standard headers for informing this.

### DIFF
--- a/copyparty/httpcli.py
+++ b/copyparty/httpcli.py
@@ -94,6 +94,7 @@ class HttpCli(object):
     def run(self):
         """returns true if connection can be reused"""
         self.keepalive = False
+        self.is_https = False
         self.headers = {}
         self.hint = None
         try:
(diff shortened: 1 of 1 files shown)


9001:
ayy nice catch :+1: 
but this'll probably be useful other places too, how about making it `self.is_https` and putting it next to `self.keepalive`?
oh and good idea to lowercase it before comparing!

Daedren:
> ayy nice catch 👍 but this'll probably be useful other places too, how about making it `self.is_https` and putting it next to `self.keepalive`? oh and good idea to lowercase it before comparing!

Yeah I agree.
I've now made those fixes in the newest update.

9001:
good stuff :ok_hand: 
thx for partying!

PR #5, opened 2021-10-31
contrib: add freebsd rc script

kipukun:
since im a fuckin hipster


9001:
:ok_hand: :ok_hand: :ok_hand:

9001:
fug i keep doing that

PR #12, opened 2022-09-11
Use grids for the grid (if supported by browser)

epertt:
demo @ you know where

1. grid items centered if sidebar is off
2. spaced evenly if sidebar is on
3. aligned left if there are too few items to fill a row (regardless of sidebar visibility)

i can't read your JS so idk if i did something dumb there. maybe you can figure out something better

### DIFF
--- a/copyparty/web/browser.css
+++ b/copyparty/web/browser.css
@@ -2893,3 +2893,12 @@ html.d #treepar {
 		margin-top: 1.7em;
 	}
 }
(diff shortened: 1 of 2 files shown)


icxes:
should probably remember to check which account i'm on before doing that

9001:
looks dope :+1: new css tricks ayy

and dont worry about the js, not like it could possibly get any worse w

thx!

PR #15, opened 2023-02-02
add PKGBUILD

icxes:
Tested and works on my machine but you'll probably want to test in a VM or something, some stuff might be a bit sus since it's my first PKGBUILD, should be pretty obvious to you if something's fucked just by looking at it tho

### DIFF
--- a/contrib/package/arch/PKGBUILD
+++ b/contrib/package/arch/PKGBUILD
@@ -15,25 +15,25 @@ source=("git+https://github.com/9001/${_pkgname}.git")
 sha256sums=("SKIP")
 
 prepare() {
-	cd "${srcdir}/${_pkgname}"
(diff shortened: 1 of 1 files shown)


9001:
lmao i love the seds this is perfect
thx

icxes:
just noticed it's actually broken, gimme 5

9001:
oh this might be a good excuse to try out config file includes actually, ok so if the pkgbuild were to install `/etc/copyparty.conf` with the contents `% copyparty.d` and also creates an empty folder named `/etc/copyparty.d` it should try to load all *.conf files from there

might be a better alternative to modifying the systemd script, could just add `-c /etc/copyparty.conf` to load that

icxes:
i'll PR with just the fixed one for now and look at doing what you suggested later, sounds dope tho

PR #22, opened 2023-04-04
Nix/NixOS support

chinponya:
This PR adds:
- a Nix package
- a NixOS module
- simple documentation on how to use these
- a script for easily updating the package

The package is built using the sfx release. It adds most of the necessary dependencies.
Other scripts such as up2k are not included, but this could be done in the future.
It's intentionally omitting Pillow, since it looks like all the extensions are handled by either vips or ffmpeg, which are included.
I did not bother with support for vamp/vamp-sdk and related memes. This is something I could do eventually, but I've suffered enough for now.

The NixOS module is running copyparty in a restricted environment by default, kinda like prisonparty, but using systemd features around namespaces instead of a chroot. 
I did manually test the features that looked like they could break and everything seems to work.
Only additional requirement is that the user has to specify read/write paths in the configuration. This could be automated by specifying the copyparty configuration as a nix module and serializing the real config from it. This would give us the ability to read parts of the config and deriving which paths should be readable/writable.
Given that the config format used is not compliant with any existing format that I know of, I did not bother doing that.
Another quirk, which is more on the copyparty to fix, is that the certificates live in `$XDG_CONFIG_HOME`, with no ability to configure it. That means that unless the user manually modifies the certificates in `/var/lib/copyparty`, it will use a self-signed certificate. This could be worked around in the NixOS module, but it would be a lot less hacky, if the cert path was configurable.

I've also included a script for updating the Nix package in `contrib/packages/nix/copyparty/update.py`.
It will download the latest sfx.py from github and put its version and hash in `pin.json`, which is then used by the Nix derivation.
I then immediately realized that this mode of operation sucks for maintainers, because you won't have a release when it would make sense to run the script, which is just before adding a tag.
That's why you can also give `update.py` a path to an sfx file and it will update `pin.json` with a version contained within. That might break because I have some shitty regex doing it, but it should be good enough.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,6 @@ up.*.txt
 .hist/
 scripts/docker/*.out
 scripts/docker/*.err
+
+# nix build output link
+result
--- a/README.md
+++ b/README.md
@@ -1138,32 +1138,62 @@ for this setup, you will need a [flake-enabled](https://nixos.wiki/wiki/Flakes) 
 }
 ```
 
-copyparty on NixOS is configured via `services.copyparty.config`, for example:
+copyparty on NixOS is configured via `services.copyparty` options, for example:
 ```nix
 services.copyparty = {
   enable = true;
-  config = ''
-    [global]
-      i: 0.0.0.0
-      no-reload
+  # directly maps to values in the [global] section of the copyparty config.
+  # see `copyparty --help` for available options
+  settings = {
+    i = "0.0.0.0";
+    # use lists to set multiple values
+    p = [ 3210 3211 ];
+    # use booleans to set binary flags
+    no-reload = true;
+    # using 'false' will do nothing and omit the value when generating a config
+    ignored-flag = false;
+  };
 
-    # create users
-    [accounts]
-      # username: password
-      ed: 123
+  # create users
+  accounts = {
+    # specify the account name as the key
+    ed = {
+      # provide the path to a file containing the password, keeping it out of /nix/store
+      # must be readable by the copyparty service user
+      passwordFile = "/run/keys/copyparty/ed_password";
+    };
+    # or do both in one go
+    k.passwordFile = "/run/keys/copyparty/k_password";
+  };
 
-    # create a volume
-    [/]               # create a volume at "/" (the webroot), which will
-      /srv/copyparty  # share the contents of "/srv/copyparty"
-      accs:
-        r: *          # everyone gets read-access, but
-        rw: ed        # the user "ed" gets read-write
-  '';
-  # the service runs in an isolated environment by default
-  # any directory you reference in the volume configuration
-  # needs to be added here, in order to make it discoverable
-  # with the exception of /var/lib/copyparty, which is always available
-  readWritePaths = [ "/srv/copyparty" ];
+  # create a volume
+  volumes = {
+    # create a volume at "/" (the webroot), which will
+    "/" = {
+      # share the contents of "/srv/copyparty"
+      path = "/srv/copyparty";
+      # see `copyparty --help-accounts` for available options
+      access = {
+        # everyone gets read-access, but
+        r = "*";
+        # users "ed" and "k" get read-write
+        rw = [ "ed" "k" ];
+      };
+      # see `copyparty --help-flags` for available options
+      flags = {
+        # "fk" enables filekeys (necessary for upget permission) (4 chars long)
+        fk = 4;
+        # scan for new files every 60sec
+        scan = 60;
+        # volflag "e2d" enables the uploads database
+        e2d = true;
+        # "d2t" disables multimedia parsers (in case the uploads are malicious)
+        d2t = true;
+        # skips hashing file contents if path matches *.iso
+        nohash = "\.iso$";
+      };
+    };
+  };
   # you may increase the open file limit for the process
   openFilesLimit = 8192;
 };
--- a/README.md
+++ b/README.md
@@ -99,6 +99,8 @@ just run **[copyparty-sfx.py](https://github.com/9001/copyparty/releases/latest/
 
 * or install through pypi (python3 only): `python3 -m pip install --user -U copyparty`
 * or if you cannot install python, you can use [copyparty.exe](#copypartyexe) instead
+* or install through nix: `nix profile install github:9001/copyparty`
+  * requires a [flake-enabled](https://nixos.wiki/wiki/Flakes) installation of nix
 * or if you are on android, [install copyparty in termux](#install-on-android)
 * or if you prefer to [use docker](./scripts/docker/) 🐋 you can do that too
   * docker has all deps built-in, so skip this step:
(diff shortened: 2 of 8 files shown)


9001:
Oh boy, thanks! Looks awesome :+1: but never had an excuse to try nix before so I'll take a proper look at this in a few weeks :pray: 

just going off the PR description it looks solid -- good decisions. Sorry about the crazy config syntax, how does nixos usually do configs? And good call regarding XDG_CONFIG_HOME, will definitely make that configurable now that there's finally a reason to do that.

and yep, should probably take a proper look at the release procedure sometime soon -- it was nicely automated until docker containers + the archpkg was added, so now it's something like... add tag, build sfx (since it reads the tag (of course)), make pypi release, update archpkg version, make docker packages (which might fail due what i think is a podman-multiarch race condition), and finally push tags. Wew. So thx for the motivation to get all that cleaned up o/

chinponya:
There is nothing wrong with the nixos config taking a multi-line string. Many packages work that way, but for some time now, nixpkgs is switching to nix options where possible.
This is done in one of two ways:
1. Define every single config option in terms of nix options, along with a type, default, description and an example. This is the best way to do it, because it gives us better build time guarantees: it will only accept configuration in a specific shape and type. These options could be generated from some sort of schema, if one existed.
2. The config option could be defined as a free-form type, so that it can be written in nix syntax. 
This is actually how most modules for packages that target a config file are made.

For either solution, the nix options are then serialized to a format a program can read. Serializers for some common formats already exist, so this is typically quite easy.

It shouldn't be too difficult to support a brand new format. The main obstacle is that it may need to be written in pure nix, which can be more challenging. There's also the fact that I do not know how the current copyparty config format works exactly.

Aside from the advantage of the nix configuration being nicer and more likely to catch errors early, when implemented in terms of nix options, there's actually one slightly concerning issue.
Everything built by nix lives in the /nix/store, which is world-readable. That includes your user passwords that get placed in the config.
The easiest way to solve this, on the nix end, is to supply the passwords via a file. For this, the program needs to support reading secrets from a file.
There are workarounds, of course. Secrets could be omitted from the main config and instead would be provided as paths in a separate nix option. The nix option containing the secrets wouldn't be used when serializing the config, but instead supplied in some other way in the systemd ExecStart.
The solution could be implemented in a nicer way if the config wasn't a one large string, but a nix option. This way, the module could treat certain values in a special way without making the module's API a mess.
However, the question remains: do we actually give a shit?


Also, just to clarify: when I mentioned that the certs are stored in XDG_CONFIG_HOME, I don't mean that the path of a directory it puts there should be configurable - it's just the certificate path that needs it.

chinponya:
In other words, right now we do this:
```nix
services.copyparty = {
  enable = true;
  config = ''
    [global]
      i: 0.0.0.0
      no-reload

    [accounts]
      ed: 123  # world-readable

    [/]
      /srv/copyparty  # needs to be added to copyparty.readWritePaths to work
      accs:
        r: *
        rw: ed
  '';
  readWritePaths = [ "/srv/copyparty" ];
};
```

But in the ideal world, we could be doing something like this:
```nix
services.copyparty = {
  enable = true;
  config =  {
    global = {
      i = "0.0.0.0";
      options = [ "no-reload" ];
    };

    accounts = {
      ed = {
        # would be serialized as a path in the actual config and copyparty would just read the file
        passwordFile = "/run/copyparty/ed_password"; # contains "123"
      };
    };

    volumes = {
      "/" = {
        # gets automatically placed in readWritePaths,
        # because nix can introspect the config
        path = "/srv/copyparty";
        accs = {
          all = "r";
          ed = "rw";
        };
      };
    };
  };
};
```

Or at least this:
```nix
services.copyparty = {
  enable = true;
  # needs to exist as a separate option, so that nix can introspect it
  # and do the right thing, avoiding placing secrets in a world-readable config
  accounts = {
    ed = {
      passwordFile = "/run/copyparty/ed_password"; # contains "123"
    };
  };
  config = ''
    [global]
      i: 0.0.0.0
      no-reload

    [/]
      /srv/copyparty  # needs to be added to copyparty.readWritePaths to work
      accs:
        r: *
        rw: ed
  '';
  readWritePaths = [ "/srv/copyparty" ];
};
```

All that said, it is not really that important. I wouldn't PR it in the current state if I thought it was and it can always be changed.
It could be also just adding more maintenance burden. Suddenly you end up with another place that cares deeply about the shape of the config and needs to be kept up to date to work. The same kind of trap I fell into with the previous attempt at adding nix support, before nuking it all.

chinponya:
I've made some changes. Now everything is defined as nix options.
The nix config is then serialized to a copyparty config format.
It puts a version of the config with the placeholders in the `/nix/store`, which are copied by systemd in `preInit` into `/run/copyparty` and has the placeholders replaced with the contents of `passwordFile`s. When this fails, the entire unit fails, so there's no risk of accidentally setting your password to `{{password-ed}}`.
It also automatically bind mounts `passwordFile`s as readable and all volume paths as read/write.

I think this clears up most of the issues, at the cost of increasing the complexity of the module somewhat.
One caveat is that I keep the nix option definitions to a bare minimum, so it is possible to write invalid config this way, which will only fail at the unit start. Fixing that would require listing every single valid option in the nix module and keeping that in sync with copyparty, which I don't think is really worth it.

Now it looks like this:
```nix
services.copyparty = {
  enable = true;
  # directly maps to values in the [global] section of the copyparty config.
  # see `copyparty --help` for available options
  settings = {
    i = "0.0.0.0";
    # use lists to set multiple values
    p = [ 3210 3211 ];
    # use booleans to set binary flags
    no-reload = true;
    # using 'false' will do nothing and omit the value when generating a config
    ignored-flag = false;
  };

  # create users
  accounts = {
    # specify the account name as the key
    ed = {
      # provide the path to a file containing the password, keeping it out of /nix/store
      # must be readable by the copyparty service user
      passwordFile = "/run/keys/copyparty/ed_password";
    };
    # or do both in one go
    k.passwordFile = "/run/keys/copyparty/k_password";
  };

  # create a volume
  volumes = {
    # create a volume at "/" (the webroot), which will
    "/" = {
      # share the contents of "/srv/copyparty"
      path = "/srv/copyparty";
      # see `copyparty --help-accounts` for available options
      access = {
        # everyone gets read-access, but
        r = "*";
        # users "ed" and "k" get read-write
        rw = [ "ed" "k" ];
      };
      # see `copyparty --help-flags` for available options
      flags = {
        # "fk" enables filekeys (necessary for upget permission) (4 chars long)
        fk = 4;
        # scan for new files every 60sec
        scan = 60;
        # volflag "e2d" enables the uploads database
        e2d = true;
        # "d2t" disables multimedia parsers (in case the uploads are malicious)
        d2t = true;
        # skips hashing file contents if path matches *.iso
        nohash = "\.iso$";
      };
    };
  };
  # you may increase the open file limit for the process
  openFilesLimit = 8192;
};
```

9001:
Finally got this tested, works great! Thanks a lot for the contribution o/

PR #50, opened 2023-08-09
Update wget.py to prevent shell injection

mvsite:
used urlparse and shlex.quote to limit user to intended capability

### DIFF
--- a/bin/hooks/wget.py
+++ b/bin/hooks/wget.py
@@ -4,7 +4,8 @@ import os
 import sys
 import json
 import subprocess as sp
-
+from urllib.parse import urlparse
+from shlex import quote
 
 _ = r"""
 use copyparty as a file downloader by POSTing URLs as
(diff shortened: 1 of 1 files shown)


mvsite:
I realize this is very low priority and only a nitpick about capability for authorized users, *and* I didn't actually test to see if it was exploitable in the first place, but honestly I am just testing github and haven't really used it before (as a contributor).  Thanks for making copyparty!

mvsite:
..and in the same day I learned that you can't delete a pull request once it is created.  That's too bad, because I'm pretty sure this wasn't exploitable anyway. :)  Oh well.  Have a nice day!

9001:
No worries, thanks for the concern (and for using copyparty!) 👍

You're right -- this would have been necessary if the `subprocess.check_call` had been invoked with `shell=True`. But thanks to the `--` (stop argument parsing) and the `shell=False` (its default value), the worst-case outcome of any malicious input should be that wget simply fails safely.

But I like the idea of a protocol whitelist, since this could avoid any creative / unintended uses of wget. For example, wget does not currently support the `file` protocol, but we definitely don't want anyone to be able to use that! So I'd like to merge this with just a few changes if you're up for it :>

* quoting the url is not necessary (and could potentially cause issues), since this is handled by subprocess already
* the protocol whitelist should also allow `ftp`

PR #94, opened 2024-08-21
Add support for chinese

ultwcz:
Hi， I added some translation for Chinese.

ultwcz,
Best regargs.


9001:
awesome, thanks a lot! Looking forward to merging this --

I wish I could read chinese so I could help proofread, but I trust that you've done a good job and managed to capture all the details :>

however, there are a few things that must be fixed -- due to my lazy/crazy mixing of html-tags and i18n text, we've hit some syntax errors...

for the following strings, please use `$N` instead of `\n` as line separator:

ot_u2i, ot_u2w, wt_ren, wt_del, wt_cut, wt_pst, wt_selall, wt_seldl, wt_grid, wt_prev, wt_play, wt_next, ut_mt, ut_par, ul_btn, cut_turbo, cut_flag, cut_az, cut_mt, cdt_ask, tt_entree, tt_detree, tt_ftree, tt_hover, mt_prescan, mt_fullpre, mt_waves, mt_oseek, mt_eq, mt_drc, ft_paste, tvt_close, tvt_dl, tvt_prev, tvt_next, tvt_sel, tvt_edit

...but please remember that `\n` and `<em>` is correct in some of the strings! so don't just replace every `\n` with `$N`, it depends where and how the string is used in the app :>

for example `\n` is correct in mm_opusen and ft_paste.

I have done a review on github as well, to cover some more specific situations, so please take a look there as well.

Your PR made me realize that we need a better way to perform translation, and to compare the translation to english. I hope to make a small program for this. But since there is no utility for that right now, I quickly made the following unix command to investigate the issues... I'll include it here, just in case you might also find it useful:

```bash
cat copyparty/web/browser.js | awk '/^\t\}/{fa=0;fb=0} !/":/{next} /"eng":/{fa=1} /"chi":/{fb=1} fa{a[ia++]=$0} fb{b[ib++]=$0} END{for (i=0;i<ia;i++) printf "%s\n%s\n\n",a[i],b[i]}' | sed -r $'s/\\\\/\033[1;37;41m\\\\\033[0m/g; s/\$N/\033[1;37;45m$N\033[0m/g; s/([{}])/\033[34m\\1\033[0m/g; s/"/\033[44m"\033[0m/g; s/\'/\033[45m\'\033[0m/g; s/&/\033[1;43;30m&\033[0m/g' | sed -r 's/^\t+//' | less -R
```

thanks again, and looking forward to testing the next changeset :grin:

ultwcz:
Okay, I found my changs is very The method I used was very simple and crude, and I translated directly. Of course, I also noticed that you used HTML statements and tags directly mixed, and the use of single quotes and double quotes was somewhat different. The unix command you mentioned helped me a lot. I will continue to optimize it and look forward to merging it into your main line. Thank you very much. I am very touched and honored for this translation work.

I would also like to ask you how you maintain so many js codes. I admire you very much.

Thank you

ultwcz:
Thanks for your feedback! I've updated the code as you suggested. Please check it out and see if it works as you wanted. If there's anything else I can improve, please let me know.

9001:
Awesome, this is looking pretty good now :grin:  :+1:

> I would also like to ask you how you maintain so many js codes

Eh, I just think this is easier than using many dependencies -- then it quickly becomes hard to see exactly why things are happening hehe

Thank you again for the contribution, and don't be afraid to open another PR if you happen to notice any inaccuracies at some point :>

PR #95, opened 2024-08-23
Chinese ranslation fix

ultwcz:
Hi，

I am Using the script to correct the improper use of "" and replace it with ''. It is great!!!.

Also, check the newly added Chinese translation.

 By the way, compared to your work, mine seems quite rough.


9001:
Nice, thanks for the translation fixes! That was fast hehe

Regarding `"` vs`'` for strings, I have some good news - we don't really have to worry about those. The javascript doesn't care, since they mean the same thing. But the script is pretty dumb, and just sees that there's  difference, so it complains :-) 

 
But it's fine either way, and your changes will make the script complain less, so that's nice :grin: :+1: 

BTW, I am adding one more string -- what should the following be in Chinese?

English: you cannot select more than one folder, nor can you mix flies and folders in one selection
Norwegian: du kan ikke markere flere mapper samtidig, eller kombinere mapper og filer

These are regarding the sharing feature, where you select the files and folders you want to share. It is OK to select either nothing, or one folder, or one or more files. This message will be displayed if you select more than one folder. Just to avoid any confusion, the meaning is that it is fine to share as many folders as you'd like, but then you need to perform the steps to create a share repeatedly, doing one folder at a time. This was a bit tricky to get right in Norwegian, so I wanted to point it out just in case.

Thanks, and have a nice weekend!

ultwcz:
> Nice, thanks for the translation fixes! That was fast hehe
> 
> Regarding `"` vs`'` for strings, I have some good news - we don't really have to worry about those. The javascript doesn't care, since they mean the same thing. But the script is pretty dumb, and just sees that there's difference, so it complains :-)
> 
> But it's fine either way, and your changes will make the script complain less, so that's nice 😁 👍
> 
> BTW, I am adding one more string -- what should the following be in Chinese?
> 
> English: you cannot select more than one folder, nor can you mix flies and folders in one selection Norwegian: du kan ikke markere flere mapper samtidig, eller kombinere mapper og filer
> 
> These are regarding the sharing feature, where you select the files and folders you want to share. It is OK to select either nothing, or one folder, or one or more files. This message will be displayed if you select more than one folder. Just to avoid any confusion, the meaning is that it is fine to share as many folders as you'd like, but then you need to perform the steps to create a share repeatedly, doing one folder at a time. This was a bit tricky to get right in Norwegian, so I wanted to point it out just in case.
> 
> Thanks, and have a nice weekend!

Hi, @9001 ,
I translated it following, refering english:
English: you cannot select more than one folder, nor can you mix flies and folders in one selection
Norwegian: du kan ikke markere flere mapper samtidig, eller kombinere mapper og filer
Chinese: 你不能同时选择多个文件夹，也不能同时选择文件夹和文件

Thanks

PR #98, opened 2024-09-17
RFC: Python 3.7 package resources support

shizmob:
This series of patches adds support for loading web resources through Python's `importlib.resources` abstraction, allowing them to be packed into whatever module loading mechanism is used (e.g. `zipimport`). This would allow distributing copyparty as a single `.py` without needing any kind of file system access in `/tmp` or elsewhere.

Three pending improvements that make it RFC for now:
- ~~Make package name used for access configurable through `E` instead of hardcoded in `util.py`~~
- Verify `walk_resources()` correctness
- ~~Add support for the older [`pkg_resources`](https://setuptools.pypa.io/en/latest/pkg_resources.html) API~~

### DIFF
--- a/copyparty/__init__.py
+++ b/copyparty/__init__.py
@@ -54,6 +54,7 @@ except:
 
 class EnvParams(object):
     def __init__(self) -> None:
+        self.pkg = ""
         self.t0 = time.time()
         self.mod = ""
         self.cfg = ""
--- a/copyparty/__init__.py
+++ b/copyparty/__init__.py
@@ -54,7 +54,7 @@ except:
 
 class EnvParams(object):
     def __init__(self) -> None:
-        self.pkg = ""
+        self.pkg = None
         self.t0 = time.time()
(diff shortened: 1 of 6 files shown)


shizmob:
Motivating example: attached is a version of copyparty (`tar`'d because GitHub, from the HEAD of my branch at time of writing, https://github.com/9001/copyparty/pull/98/commits/2997c427aadc02038f41df77a1612d891e624fcd) packed with a simple packer I wrote into a single `.py` file that doesn't need file system access other than for serving files.

[copyparty.py.tar.gz](https://github.com/user-attachments/files/17026528/copyparty.py.tar.gz)

9001:
Nice! This will be great for the pyz / zipapp too, since that currently extracts the webdeps and jinja-templates into a temporary folder, which was somewhat problematic due to systemd's tmpfiles cleaner.

Thanks for the cool stuff :grin:

PR #118, opened 2024-11-22
Allow multiple CIDR ranges in `build_netmap` when using LAN shorthands

codemicro:
This PR makes it possible to use LAN shorthands when specifying custom CIDR ranges. For example, in the situation where you want to allow connections from both the various local networking places and from Tailscale, you would want to do:

```
xff-src: lan, 100.64.0.0/10
```

but this currently does not work and instead causes copyparty to fail to start.

This is because shorthands are currently detected by comparison to the whole comma-separated string - this PR moves shorthand expansion to after the parsing step to make the above example work as expected.

:)

### DIFF
--- a/copyparty/util.py
+++ b/copyparty/util.py
@@ -2709,12 +2709,21 @@ def build_netmap(csv: str, defer_mutex: bool = False):
     if csv in ("any", "all", "no", ",", ""):
         return None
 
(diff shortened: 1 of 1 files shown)


9001:
Nice! thanks for catching and fixing this :grin: :+1:

PR #135, opened 2025-02-06
Update README.md

coderofsalvation:
added sandboxing example via bubblewrap (+added detail dropdown)

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,9 @@ just run **[copyparty-sfx.py](https://github.com/9001/copyparty/releases/latest/
 * or if you are on android, [install copyparty in termux](#install-on-android)
 * or maybe you have a [synology nas / dsm](./docs/synology-dsm.md)
 * or if your computer is messed up and nothing else works, [try the pyz](#zipapp)
+* or if you don't trust copyparty yet and want to isolate it a little, then...
+  * ...maybe [prisonparty](./bin/prisonparty.sh) to create a tiny [chroot](https://wiki.archlinux.org/title/Chroot) (very portable),
+  * ...or [bubbleparty](./bin/bubbleparty.sh) to wrap it in [bubblewrap](https://github.com/containers/bubblewrap) (much better)
 * or if you prefer to [use docker](./scripts/docker/) 🐋 you can do that too
   * docker has all deps built-in, so skip this step:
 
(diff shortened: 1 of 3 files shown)


9001:
sweet! but hold on, I've got an idea...

so currently there is [prisonparty](https://github.com/9001/copyparty/tree/hovudstraum/bin#prisonpartysh) -- an old-fashioned, jankier alternative which does mostly the same thing, just not as well as bubblewrap does. I still want to keep that around since bubblewrap isn't always available (but `/sbin/chroot` always is)... but it would be great to have a bubblewrap counterpart to this.

so how about making this a wrapper-script similar to prisonparty.sh, but just bubblewrap-based instead of chroot? could call it bubbleparty :grin: (or maybe you have a better idea!)

wouldn't need to be fancy or anything, could be exactly what you suggested in the readme already but just as a script instead... except, maybe with the 1st argument being the path to `copyparty-sfx.py` for ease of use? and it would be cool if it would forward any commandline args to copyparty, and come to think of it... maybe this would work?

```patch
-./copyparty-sfx.py --no-crt -c cfg/config.conf \
+"$@" \
```

as for the readmes, how about...

```patch
diff --git a/README.md b/README.md
index 54f9521f..0d7b85e5 100644
--- a/README.md
+++ b/README.md
@@ -145,4 +145,7 @@ just run **[copyparty-sfx.py](https://github.com/9001/copyparty/releases/latest/
 * or maybe you have a [synology nas / dsm](./docs/synology-dsm.md)
 * or if your computer is messed up and nothing else works, [try the pyz](#zipapp)
+* or if you don't trust copyparty yet and want to isolate it a little, then...
+  * ...maybe [prisonparty](./bin/prisonparty.sh) to create a tiny [chroot](https://wiki.archlinux.org/title/Chroot) (very portable),
+  * ...or [bubbleparty](./bin/bubbleparty.sh) to wrap it in [bubblewrap](https://github.com/containers/bubblewrap) (much better)
 * or if you prefer to [use docker](./scripts/docker/) 🐋 you can do that too
   * docker has all deps built-in, so skip this step:
@@ -1187,5 +1190,5 @@ dependencies: `python3 -m pip install --user -U impacket==0.11.0`
 some **BIG WARNINGS** specific to SMB/CIFS, in decreasing importance:
 * not entirely confident that read-only is read-only
-* the smb backend is not fully integrated with vfs, meaning there could be security issues (path traversal). Please use `--smb-port` (see below) and [prisonparty](./bin/prisonparty.sh)
+* the smb backend is not fully integrated with vfs, meaning there could be security issues (path traversal). Please use `--smb-port` (see below) and [prisonparty](./bin/prisonparty.sh) or [bubbleparty](./bin/bubbleparty.sh)
   * account passwords work per-volume as expected, and so does account permissions (read/write/move/delete), but `--smbw` must be given to allow write-access from smb
   * [shadowing](#shadowing) probably works as expected but no guarantees
diff --git a/bin/README.md b/bin/README.md
index 495292a7..6acae7d6 100644
--- a/bin/README.md
+++ b/bin/README.md
@@ -79,2 +79,7 @@ cd /mnt/nas/music/.hist
 * run copyparty in a chroot, preventing any accidental file access
 * creates bindmounts for /bin, /lib, and so on, see `sysdirs=`
+
+
+
+# [`bubbleparty.sh`](bubbleparty.sh)
+* run copyparty with [bubblewrap](https://github.com/containers/bubblewrap), like prisonparty but better
```

what do you think? :>

EDIT: oh btw, should mention I'm very much not a fan of collapsible sections in readmes, since I find that it hurts discoverability / ctrl-f... but maybe that's just me hehe

coderofsalvation:
Oh I totally missed prisonparty :)
And yes, lets ditch the collapse-sections and add a separate script.

I've updated the commit of this PR: https://github.com/9001/copyparty/commit/cef50638cd4844670175ac16e666f3572bdf2462

9001:
Awesome stuff, thanks o/

PR #140, opened 2025-02-26
Silence missing mainProgram in Nix package

daimond113:
Previously, when using the Nix package warnings such as:
```
evaluation warning: getExe: Package "copyparty-1.16.15" does not have the meta.mainProgram attribute. We'll assume that the main program has the same name for now, but this behavior is deprecated, because it leads to surprising errors when the assumption does not hold. If the package has a main program, please set `meta.mainProgram` in its definition to make this warning go away. Otherwise, if the package does not have a main program, or if you don't control its definition, use getExe' to specify the name to the program, such as lib.getExe' foo "bar".
```
were emitted, so to silence them this PR adds the `meta.mainProgram` field.


9001:
nice, thanks for the fix 👍

PR #152, opened 2025-04-06
Apply custom fonts to up2k parallel uploads input

thaddeuskkr:
Currently, the parallel uploads input in the up2k configuration table doesn't use the custom font set in `:root`, as it doesn't inherit from `html`.

On my instance of copyparty, `Tahoma` is used instead on Firefox, and `Arial` on Edge (I assume default sans-serif fonts?) - so this PR fixes that.

### DIFF
--- a/copyparty/web/browser.css
+++ b/copyparty/web/browser.css
@@ -2631,7 +2631,6 @@ html.y #bbox-overlay figcaption a {
 	color: var(--fg-max);
 	background: var(--u2-txt-bg);
 	border: 1px solid #777;
-	font-family: var(--font-main), sans-serif;
 	font-size: 1.2em;
 	padding: .15em 0;
 	height: 1.05em;
--- a/copyparty/web/browser.css
+++ b/copyparty/web/browser.css
@@ -2631,6 +2631,7 @@ html.y #bbox-overlay figcaption a {
 	color: var(--fg-max);
 	background: var(--u2-txt-bg);
 	border: 1px solid #777;
+	font-family: var(--font-main), sans-serif;
 	font-size: 1.2em;
 	padding: .15em 0;
 	height: 1.05em;
--- a/copyparty/web/ui.css
+++ b/copyparty/web/ui.css
@@ -381,6 +381,9 @@ html.y .btn:focus {
 	box-shadow: 0 .1em .2em #037 inset;
 	outline: #037 solid .1em;
 }
+input, button {
+	font-family: var(--font-main), sans-serif;
+}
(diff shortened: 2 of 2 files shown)


9001:
Nice, thanks! Turns out the same issue also applies to every other textbox, and also buttons (for example in the batch-rename UI)... Do you think it would be a good idea to apply this to all of them? Something like this in ui.css maybe:

```patch
--- a/copyparty/web/ui.css
+++ b/copyparty/web/ui.css
@@ -380,8 +380,11 @@ html.y #pctl *:focus,
 html.y .btn:focus {
        box-shadow: 0 .1em .2em #037 inset;
        outline: #037 solid .1em;
 }
+input, button {
+       font-family: var(--font-main), sans-serif;
+}
 input[type="submit"] {
        cursor: pointer;
 }
```

Screenshot_2025-04-06_08-24-37
Screenshot_2025-04-06_08-24-54
Screenshot_2025-04-06_08-40-11
Screenshot_2025-04-06_08-39-49

Trying to think of any reasons that this could cause any problems, but it should be safe I think :+1:

9001:
Or... wait, I just remembered this -- https://github.com/9001/copyparty/blob/hovudstraum/docs/rice/README.md

> and because textboxes don't inherit fonts by default, you can force it like this:  
> `input[type=text], input[type=submit], input[type=button] { font-family: var(--font-main) }`

maybe, since the browser doesn't do font-inherit by default, it would be best to keep the default of not overriding the input fonts, and instead make it possible to override it in this way if you want to?

in other words, maybe we shouldn't change any of this after all heh

thaddeuskkr:
> maybe, since the browser doesn't do font-inherit by default, it would be best to keep the default of not overriding the input fonts, and instead make it possible to override it in this way if you want to?
> 
> in other words, maybe we shouldn't change any of this after all heh

i get where you're coming from yeah, but just curious - why not just override the browser's defaults? imo having a unified font throughout, even if its not a custom font set by the user, is good

especially since it's just changing the font so it shouldn't affect accessibility or anything right?

sflr btw i need to turn on notifs, do ping or dm me on discord if u can't reach me

9001:
> sflr btw i need to turn on notifs

w no worries 👌

only thing I'm worried about is that there must be a reason that it's not the default, maybe due to some particular OS with some particular language or something... but I'm probably overthinking it ¯⁠\⁠_⁠(⁠ツ⁠)⁠_⁠/⁠¯

yeah let's go with always setting the font for input and button in ui.css 👍

PR #153, opened 2025-04-11
full lowercase on login button to match the rest of the control panel

thaddeuskkr:
the rest of the control panel page is basically all lowercase, so i see no reason the login button shouldn't be, imo it looks better and more unified with no capital L

soz for the small changes per PR :)

image

### DIFF
--- a/copyparty/web/splash.html
+++ b/copyparty/web/splash.html
@@ -122,7 +122,7 @@
 				<input type="hidden" id="la" name="act" value="login" />
 				<input type="password" id="lp" name="cppwd" placeholder=" password" />
 				<input type="hidden" name="uhash" id="uhash" value="x" />
-				<input type="submit" id="ls" value="Login" />
+				<input type="submit" id="ls" value="login" />
 				{% if chpw %}
 				<a id="x" href="#">change password</a>
 				{% endif %}


9001:
heh, yeah my concern was that someone might misread it as "iogin" and think "what's a iogin??"

like how everyone thought [ievan polkka](https://www.youtube.com/watch?v=hqthspSKZV8) was "Levan polkka" :p

it's nice how your font makes it obvious, but the default one on windows and linux doesn't... ahh

thaddeuskkr:
> heh, yeah my concern was that someone might misread it as "iogin" and think "what's a iogin??"

imo it's implied easily by the password field beside it

> it's nice how your font makes it obvious, but the default one on windows and linux doesn't... ahh

maybe we could use a better default font? 

...nah im getting ahead of myself

9001:
> maybe we could use a better default font?

it's tempting right? :p but I've noticed that some OS'es / browsers really struggle with rendering custom fonts decently, like the kerning and subpixel rendering gets all whack.

I made an exception for the monospace font since all the alternatices on windows are awful (especially win7), and somehow *that specific old version* of SourceCodePro seems to render perfectly everywhere, so that one was worth it -- but don't think it's a good idea to mess with the default serif or sans-serif :>

PR #154, opened 2025-04-12
Using systemd service templates for Arch Linux

RafaelDominiquini:
Avoid the need to edit the service everytime the package is updated!


9001:
Hey, welcome and thanks 😁👍

So I'm curious -- the systemd example that ships with the arch package is really stable, hasn't been modified in years, so the idea (and my assumption) was that you would copy the file into `/etc/systemd/system` so it would override the default/example one in `/usr/lib/systemd`, meaning you would only need to edit the service-file just that one time, and then never again until it is edited upstream. Am I wrong here?

And I'll admit my systemd knowledge is really minimal, but i thought the most common usecase for templates was to run several instances of a service, and not for general configuration such as setting basic config like username/group -- https://fedoramagazine.org/systemd-template-unit-files/

I know there are some copyparty deployments where people do run several instances, and in that sense it would be nice being able to easily start several instances which would load different config files, AND run with their own separate XDG_CONFIG_HOME locations (since those cannot be shared between instances), and templates would help with that. Even so, it looks like people tend to use the same user/group for all of their instances, and only changing the folder it reads the `.conf` files from, plus the XDG_CONFIG_HOME. And either way it also doesn't look like that's what we're doing in this PR at all, so help me understand :>

I'll ping @icxes since he's the arch maintainer and may have some ideas too.

RafaelDominiquini:
I suggested this change because I don't like editing systemd files every time there is an update.
I don't have enough experience to say that the way I'm proposing is ideal, but I've seen several applications doing it this way.
I just made another change to avoid breaking old installations. Including systemd files in 'backups' at least prevents every update from overwriting local modifications.

Thanks

9001:
oh good, I think I see what the misunderstanding is -- you are never supposed to edit `/usr/lib/systemd/system/copyparty.service` at all; the intention is to first copy it into `/etc/systemd/system/` and then edit that new copy. That way systemd will entirely ignore the original version that's in /usr/lib/systemd, and instead use your edited copy in /etc/systemd/system.

so like this:

```bash
cp /usr/lib/systemd/system/copyparty.service /etc/systemd/system/
vim /etc/systemd/system/copyparty.service
systemctl daemon-reload
systemctl restart copyparty
```

have you tried this? does that not work? please remember the daemon-reload after editing the file, that's important!

icxes:
So, I was thinking we could probably add a user service (`/usr/lib/systemd/user/copyparty.service`) that starts copyparty as the current user and gets configuration from `$HOME/.config/copyparty/copyparty.conf` since I think(?) it should work fine with user permissions anyway, and then something like what's in this PR if you need to run it as a system service, that gets the configuration from `/etc/copyparty.d/` like it does now. That way you could just run either `systemctl enable --user copyparty` for the user service or `systemctl enable copyparty@[user]` for the system service after installation and it should Just Work. 

You could still copy the service file to `/etc/systemd/{user,system}/` if you don't want it changing without you noticing (after an update) or if you have any other reason to do so. I don't see anything breaking as a result of changing the service files since the current assumption is users already have them modified in `/etc/systemd/system/` at least.

As for people running weird setups, they would definitely have their own service files already that wouldn't be affected by changes to the package files either. There's also the option of having a user service that reads the config name from the template (so `systemctl enable --user copyparty@foo` reads `$HOME/.config/copyparty/foo.conf`, but I'm not really sure about that one? On the other hand, there's nothing preventing having both copyparty.service and copyparty@.service at the same time either...

I'd either leave this open for now or track service file improvements elsewhere.

RafaelDominiquini:
systemd service for users looks like a good idea!
I noticed here that I had only one service in the "/etc/systemd/user" folder, which I created manually.
In no other package did I need to move from the folder "/usr/lib/systemd/user" and edit it.
The same goes for system services (".../systemd/system")
My suggestion was based on the standard of use I found in other packages!
I don't have so much experience with systemd to add more to the discussion.
Thank you for considering this PR. I hope he brought at least some good ideas...

Thanks

9001:
yeah, thanks for the inspiration -- it looks like you've got icxes spinning on some ideas for where to take this at least :-)

I suspect it'll shapeshift a fair bit from here so I think the best approach is to close this PR and start from scratch... But I'll ping you in the release-notes when it hits, and try to give you some creds somehow :>

thx again!

Issue #157, opened 2025-04-22
Nixos: allow user to configure user/group copyparty runs under


gabevenberg:
Would allow better co-operation with serving a directory copyparty is not the primary owner of, such as a syncthing directory.

9001:
sounds like a good idea, but unfortunately I know absolutely nothing about nixos 😅

I'll leave this open though, in case anyone is willing to make a PR 👍

gabevenberg:
Sounds like im making a PR, then!

9001:
fixed in #159 - thanks!

PR #159, opened 2025-04-23
Nixos revamp.

gabevenberg:
Well, I started out just fixing #157, but ended up giving the whole module a bit of a cleanup. In no particular order:
* Made the formatting more consistent.
* Added explicit support for hist data to live seperate from volume data, enabled by default.
* added ability to specify user and group, to easier share the volume directories with other software such as syncthing.
* experimented with a wrapper script to easily run things like ah-cli with the same config as the production copyparty, but copyparty thwarted my efforts.


contrib/nixos/modules/copyparty.nix
```diff
@@ -269,14 +313,15 @@ in {
...
     };
 
-    users.groups.copyparty = { };
-    users.users.copyparty = {
+    users.groups.copyparty = lib.mkIf (cfg.user == "copyparty" && cfg.group == "copyparty") {};
+    users.users.copyparty = lib.mkIf (cfg.user == "copyparty" && cfg.group == "copyparty") {
       description = "Service user for copyparty";
       group = "copyparty";
-      home = home;
+      home = lib.mkIf externalStateDir;
```
chinponya: This seems to be missing an argument

contrib/nixos/modules/copyparty.nix
```diff
@@ -68,17 +70,48 @@ in {
...
+      '';
+    };
+
     openFilesLimit = mkOption {
       default = 4096;
       type = types.either types.int types.str;
       description = "Number of files to allow copyparty to open.";
     };
 
+    seperateHist = mkOption {
```
chinponya: typo? seperate != separate

contrib/nixos/modules/copyparty.nix
```diff
@@ -207,49 +240,60 @@ in {
...
-        (mapAttrsToList replaceSecretCommand cfg.accounts)}
+          (mapAttrsToList replaceSecretCommand cfg.accounts)}
       '';
 
       serviceConfig = {
         Type = "simple";
-        ExecStart = "${getExe cfg.package} -c ${runtimeConfigPath}";
+        ExecStart = ''
+          ${getExe cfg.package} -c ${runtimeConfigPath} \
+          ${optionalString (cfg.seperateHist) "--hist ${externalCacheDir}"}
```
chinponya: This is completely fine, but I liked the idea of keeping as much as possible inside a single config file. Why not merge it into `cfg.settings`?

contrib/nixos/modules/copyparty.nix
```diff
@@ -207,49 +240,60 @@ in {
...
+          ++ (mapAttrsToList (k: v: "-${v.passwordFile}") cfg.accounts);
+        BindPaths =
+          (
+            if cfg.seperateHist
+            then [externalCacheDir]
+            else []
+          )
+          ++ [externalStateDir]
+          ++ (mapAttrsToList (k: v: v.path) cfg.volumes);
+        ProtectSystem = "strict";
```
chinponya: This seems to weaken the sandbox.
Previously we used `TemporaryFileSystem = "/:ro";` which creates an empty `tmpfs`.
`ProtectSystem="strict";` only makes `/` read-only but was mutually exclusive with `TemporaryFileSystem`. Technically not an issue on systems with properly configured file perms, but I don't like tempting fate.
It's possible that I misunderstood something, somewhere.

9001:
Thanks for working on this - I don't have a lot of feedback on the nix-specific stuff, I'll leave that to the experts -- that means you two :p

but regarding [c60c8d8](https://github.com/9001/copyparty/pull/159/commits/c60c8d8e0bc744d72f895fd191325e745fc2f64c), if the idea was to launch a second instance just for password-hashing, for example with `--ah-cli` , then that should in fact be fine as a second instance next to one that's already running. Or rather, assuming the main instance is already running, then the password-hashing instance would print a harmless warning that it disabled the sessions-db and shares-db during startup, but the other way around (starting the hasher first, then main instance) would indeed break shares until the next restart... Maybe that's what you encountered? Otherwise I'm curious to hear the details :>

contrib/nixos/modules/copyparty.nix
gabevenberg: According to `man systemd.exec`, `ProtectSystem="strict"`, the entire filesystem hierarchy is mounted as read only, making it do the same thing as `TemporaryFileSystem="/:ro"`. Im mostly lifting these hardening options from other nixos modules Ive used, namely Restic and Restic-rest-server (just because Ive been looking/contributing to those recently, so they were in my shell history, lol.). If they do the same thing, id argue `ProtectSystem="strict"` is more clear.

contrib/nixos/modules/copyparty.nix
gabevenberg: Honestly, because im not sure how that would interact if the user tried to set `hist=foo` in the `settings` section. Hmm. Maybye I could just add `hist=externalCacheDir` to `cfg.settings.default`? That would complicate testing whether we need to populate `externalCacheDir`, though.

gabevenberg:
> Thanks for working on this - I don't have a lot of feedback on the nix-specific stuff, I'll leave that to the experts -- that means you two :p
> 
> but regarding [c60c8d8](https://github.com/9001/copyparty/pull/159/commits/c60c8d8e0bc744d72f895fd191325e745fc2f64c), if the idea was to launch a second instance just for password-hashing, for example with `--ah-cli` , then that should in fact be fine as a second instance next to one that's already running. Or rather, assuming the main instance is already running, then the password-hashing instance would print a harmless warning that it disabled the sessions-db and shares-db during startup, but the other way around (starting the hasher first, then main instance) would indeed break shares until the next restart... Maybe that's what you encountered? Otherwise I'm curious to hear the details :>

This is what I was getting when trying to run the `copyparty-env` script:
```
copyparty-env --ah-cli

[SFX]    this is: copyparty 1.16.21
[SFX]  packed at: 2025-04-20, 18:42:34 UTC, 1745174554
[SFX] archive is: /nix/store/spylbcjdxzh5vf5qfc4d0y6w9a4pfc6m-copyparty-1.16.21/share/copyparty-sfx.py
[SFX] python bin: /nix/store/dyqqxl67l4g1ppy618b9mwfbacln2nbv-python3-3.12.9-env/bin/python3.12
[SFX] python ver: CPython 3.12.9 (main, Feb  4 2025, 14:38:38) [GCC 14.2.1 20250322]

[SFX] found early
[SFX] jinja2: 3.1.6
[SFX] pyftpd: bundled
[SFX] sfxdir: /tmp/pe-copyparty.1000.1046187.0

copyparty v1.16.21 "COPYparty" (2025-04-20)
  CPython v3.12.9 on Linux64  [GCC 14.2.1 20250322]
   sqlite 3.48.0*1 | jinja 3.1.6 | pyftpd 1.5.10 | tftp 0.4.0

2025-04-24
14:33:59.530 root                  opening sessions-db /var/lib/copyparty/copyparty/sessions.db
14:33:59.531 root                  the sessions-db [/var/lib/copyparty/copyparty/sessions.db] is already in use by another copyparty instance (pid:1757857). This is not supported; please provide another database with --ses-db or give this copyparty-instance its entirely separate config-folder by setting another path in the XDG_CONFIG_HOME env-var. You can also disable this safeguard by setting env-var PRTY_NO_DB_LOCK=1. Will now disable sessions and instead use plaintext passwords in cookies.
14:33:59.531 root                  opening shares-db /var/lib/copyparty/copyparty/shares.db
14:33:59.531 root                  the shares-db [/var/lib/copyparty/copyparty/shares.db] is already in use by another copyparty instance (pid:1757857). This is not supported; please provide another database with --shr-db or give this copyparty-instance its entirely separate config-folder by setting another path in the XDG_CONFIG_HOME env-var. You can also disable this safeguard by setting env-var PRTY_NO_DB_LOCK=1. Will now panic.

[SFX] retcode Traceback (most recent call last):
  File "/nix/store/spylbcjdxzh5vf5qfc4d0y6w9a4pfc6m-copyparty-1.16.21/share/copyparty-sfx.py", line 277, in main
    run(tmp, j2, ftp)
  File "/nix/store/spylbcjdxzh5vf5qfc4d0y6w9a4pfc6m-copyparty-1.16.21/share/copyparty-sfx.py", line 217, in run
    run_i(ld)
  File "/nix/store/spylbcjdxzh5vf5qfc4d0y6w9a4pfc6m-copyparty-1.16.21/share/copyparty-sfx.py", line 229, in run_i
    p()
  File "/tmp/pe-copyparty.1000.1046187.0/copyparty/__main__.py", line 1812, in main
    SvcHub(al, dal, argv, "".join(printed)).run()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/pe-copyparty.1000.1046187.0/copyparty/svchub.py", line 244, in __init__
    self.setup_share_db()
  File "/tmp/pe-copyparty.1000.1046187.0/copyparty/svchub.py", line 560, in setup_share_db
    raise Exception(t)
Exception: the shares-db [/var/lib/copyparty/copyparty/shares.db] is already in use by another copyparty instance (pid:1757857). This is not supported; please provide another database with --shr-db or give this copyparty-instance its entirely separate config-folder by setting another path in the XDG_CONFIG_HOME env-var. You can also disable this safeguard by setting env-var PRTY_NO_DB_LOCK=1. Will now panic.
```

contrib/nixos/modules/copyparty.nix
```diff
@@ -49,12 +51,12 @@ let
...
   cfg = config.services.copyparty;
-  configFile = pkgs.writeText "${name}.conf" configStr;
-  runtimeConfigPath = "/run/${name}/${name}.conf";
-  home = "/var/lib/${name}";
-  defaultShareDir = "${home}/data";
+  configFile = pkgs.writeText "copyparty.conf" configStr;
+  runtimeConfigPath = "/run/copyparty/copyparty.conf";
+  externalCacheDir = "/var/cache/copyparty";
+  externalStateDir = "/var/lib/copyparty";
+  defaultShareDir = "${externalCacheDir}/data";
```
chinponya: I think this should stay in `/var/lib/copyparty/data` instead of cache, since this directory is what the default volume is configured to.
`/var/cache/copyparty/data` isn't being created anywhere either, so the service will crash unless a volume is configured.
To be fair the concept of default volume isn't needed anymore. The reason why it was there in the first place, was that copyparty would serve the current working directory as a read-only volume, which in our case was `/var/lib/copyparty`, containing a bunch of secrets.
Nowadays it goes into a fail-safe mode and doesn't serve any volumes.
That said, providing a default volume along with a directory for files to exist seems like a nicer default.

contrib/nixos/modules/copyparty.nix
```diff
@@ -207,49 +249,64 @@ in {
...
         # Hardening options
-        User = "copyparty";
-        Group = "copyparty";
-        RuntimeDirectory = name;
+        User = cfg.user;
+        Group = cfg.group;
+        RuntimeDirectory = ["copyparty"];
         RuntimeDirectoryMode = "0700";
-        StateDirectory = [ name "${name}/data" "${name}/.config" ];
+        StateDirectory = ["copyparty"];
```
chinponya, line 272: Here's where the `defaultShareDir` used to be created.

chinponya (changes requested):

contrib/nixos/modules/copyparty.nix
gabevenberg: Your right. Actually, I generalized it a bit, so that all volume dirs now get generated if they dont exist, using systemd.tmpfiles.

contrib/nixos/modules/copyparty.nix
gabevenberg: I generalized it a bit, so that all volume dirs now get generated if they dont exist, using systemd.tmpfiles.

contrib/nixos/modules/copyparty.nix
chinponya: Nice. That looks way better compared to what it was originally.

9001:
> Exception: the shares-db [/var/lib/copyparty/copyparty/shares.db] is already in use by another copyparty instance (pid:1757857). This is not supported; please provide another database with --shr-db or give this copyparty-instance its entirely separate config-folder by setting another path in the XDG_CONFIG_HOME env-var. You can also disable this safeguard by setting env-var PRTY_NO_DB_LOCK=1. Will now panic.

right, forgot that the shares-db collision is fatal, since it disables an important feature.

i think the possible workarounds are:

* making copyparty realize that it doesn't need to initialize the shares/sessions databases when entering pw-hash mode, that's the obvious approach but:
* assuming the wrapper-script is exclusively to be used for password-hashing then it would also be fine to specify PRTY_NO_DB_LOCK=1 in the script
* alternatively, the last config assignment wins , so setting `--shr ''` would also work

I'd be alright with posting a new release tonight/tomorrow with that fixed, but I recall we discussed wanting to do the password-hashing in an entirely isolated environment (with no access to the statedir), in which case the argon salt would have to be specified manually as a config argument -- would it be preferable to have both alternatives?

gabevenberg:
> right, forgot that the shares-db collision is fatal, since it disables an important feature.
> 
> i think the possible workarounds are:
> 
> * making copyparty realize that it doesn't need to initialize the shares/sessions databases when entering pw-hash mode, that's the obvious approach but:

That would probably be the most straightforward on my end, as well as probably being most useful outside of an env with a dedicated wrapper script.

> * assuming the wrapper-script is exclusively to be used for password-hashing then it would also be fine to specify PRTY_NO_DB_LOCK=1 in the script

Would work, but could be dangerous if the user tried to do something other than hasing. Could possibly hardcode the script to just do pw hashing rather than acceping any args from the user.

> * alternatively, the last config assignment wins , so setting `--shr ''` would also work

Thats probably better than the `PRTY_NO_DB_LOCK=1`.

> 
> I'd be alright with posting a new release tonight/tomorrow with that fixed, but I recall we discussed wanting to do the password-hashing in an entirely isolated environment (with no access to the statedir), in which case the argon salt would have to be specified manually as a config argument -- would it be preferable to have both alternatives?

Hmm, given that the salt is in plaintext, we could specify the salt as an argument in the script automatically, though that would run into problems where the salt would be wrong if the config also has a salt. If we do this, Id like to also have a cmd that just prints the salt copyparty is currently using.

9001:
how about a new option, `--show-ah-salt` which would print the active salt (either from statedir or args depending) before jumping into the hashing shell? 

I'll post the new ver tonight btw 👍

contrib/nixos/modules/copyparty.nix
gabevenberg: Decided to make it just a settings default, and detect whether it had been set in settings and populate that path via systemd if it has been set.

gabevenberg:
Ok, I think this is ready for final review @9001 @chinponya

chinponya (approved):
Looks good to me.
- It runs fine in a VM with or without extra configuration.
- The script works and generates valid hashes using the currently configured salt.
- I had no issues changing the user/group.

9001:
awesome, I'll go ahead and merge it then. Is there any knowledge that would get lost if i squash it, or is that fine?

chinponya:
probably the only thing would be the systemd sandboxing changes that got reverted, but that now has a big fat comment, so it should be fine to squash

9001:
sweet, thx again y'all :>

PR #164, opened 2025-05-10
Don't append .bin to unnamed files

Morganamilo:
.bin gets appended to all unnamed files uploaded via a PUT request. This happens even if a random file name is specified. .bin files are quite rare so I don't think it's a good extension to slap on to files even if they happen to be binary in nature.

Instead we should just let the file remain extensionless.

### DIFF
--- a/copyparty/httpcli.py
+++ b/copyparty/httpcli.py
@@ -2100,7 +2100,6 @@ class HttpCli(object):
         suffix = "-{:.6f}-{}".format(time.time(), self.dip())
         nameless = not fn
         if nameless:
-            suffix += ".bin"
             fn = "put" + suffix
 
         params = {"suffix": suffix, "fdir": fdir}


9001:
thanks for the suggestion! I agree it would be good to make this configurable, but I know there are deployments/usecases which assume the current filename pattern, so changing the default would be a breaking change.

I think an option such as `--put-fn` with `put-{unixtime:.6f}-{cli_ip}.bin` as the default value would be good, which would produce the current pattern by default, and would add the flexibility to specify many other alternatives as appropriate.

I'll have some free time tomorrow and could make the change then, unless you think it sounds like fun and want to give it a go, I'm fine with either :> although i wonder if it would be best to make this possible to change per-volume, and in that case the number of changes necessary to add the new option becomes a bit more intricate...

PR #171, opened 2025-06-16
Detect content-type when extension is missing or unknown

Morganamilo:
If a file has no known extension the content type gets set to application/octet-stream causing the browser try and download the file when viewed directly.

This quickly becomes annoying as many of the files I interact with often have no extension. I.e., config files, log files, LICENSE files and other random text files.

This patch uses libmagic to detect the file type and set the content-type header. It also does this for the RSS feed and webdav for sake of completeness.

This patch does not touch the front end at all so these files still have a 'txt' button and a type of '%' in the web UI. But when clicked on, the browser will display the files correctly.

This feature is enabled with the existing "magic" option. I thought this fit as the existing functionality also uses libmagic and gives file extensions to files on upload. Tell me if it should be its own option instead.

The code base was very confusing, this patch works but I have no idea if it's the way you'd like this implemented. Hopefully its acceptable as is.

### DIFF
--- a/copyparty/__main__.py
+++ b/copyparty/__main__.py
@@ -1024,7 +1024,7 @@ def add_upload(ap):
     ap2.add_argument("--u2ts", metavar="TXT", type=u, default="c", help="how to timestamp uploaded files; [\033[32mc\033[0m]=client-last-modified, [\033[32mu\033[0m]=upload-time, [\033[32mfc\033[0m]=force-c, [\033[32mfu\033[0m]=force-u (volflag=u2ts)")
     ap2.add_argument("--rand", action="store_true", help="force randomized filenames, \033[33m--nrand\033[0m chars long (volflag=rand)")
     ap2.add_argument("--nrand", metavar="NUM", type=int, default=9, help="randomized filenames length (volflag=nrand)")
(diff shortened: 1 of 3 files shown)


9001:
Thanks! This week will be a bit unpredictable so not sure when I'll get a chance to take a look but it's next on the list :>

9001:
Alright, I'll go ahead and merge this and make some adjustments afterwards --

because this currently performs mimetype detection on-demand for every time the file is requested, it introduces a severe performance degradation when the feature is enabled, reducing the webdav performance down to around 6%, making it 16x slower. But just to get this feature available for those who need it, it is good enough for now. I think I'll try to make it perform the magic detection during filesystem indexing instead, and read it from the database -- this should be way faster, maybe only dropping to 96% or so. As it is now, the performance impact from just adding this code (without enabling the feature) is just a 1% slowdown, so that is fine.

also, due to the severe performance impact, I'll separate the options into `--magic` (for uploads like before) and `--rmagic` for this new feature.

Morganamilo:
Thanks for merging. I was aware caching this info was probably better but the code is rather unnavigable with no docs and short variable names. I'm happy the general idea is accepted.

PR #175, opened 2025-06-29
Add python-magic to iv and dj docker files

Morganamilo:

This dep is needed for the magic feature. It seems to not be included in any docker files. 

I Currently build the docker containers myself as I use this feature. It would be great to have this generally available.

### DIFF
--- a/scripts/docker/Dockerfile.dj
+++ b/scripts/docker/Dockerfile.dj
@@ -15,6 +15,7 @@ RUN     apk add -U !pyc \
             py3-jinja2 py3-argon2-cffi py3-pyzmq py3-pillow \
             py3-pip py3-cffi \
             ffmpeg \
+            py3-magic \
             vips-jxl vips-heif vips-poppler vips-magick \
             py3-numpy fftw libsndfile \
             vamp-sdk vamp-sdk-libs \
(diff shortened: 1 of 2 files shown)


9001:
I think this is a good idea, but I'm confused -- in the alpine repos, the package is named `py3-magic` and not `python3-magic`, so why is this working for you?

Morganamilo:
Ah sorry I was running this patch on the djd version and did not realise the other docker images were alpine based. I've fixed the package name and build the docker image now.

9001:
Alright, that explains it -- but then I should ask, is there a reason you went with djd specifically?

The images djd djf djff dju are not intended for use; they were experiments to see whether the initial choice of using Alpine was the correct one, or whether other base-images would give better performance. The answer was no, but I kept them for reference, in case that changes in the future. But they are entirely unmaintained and have probably diverged already.

There is a warning about this in the readme but I suppose it's not very easy to notice, so I plan to make them fail to build in an obviously-intended fashion -- but if you need a non-alpine base image for some reason, then we can look at options.

Morganamilo:
I went dwon a rabbit hole trying to get random file names with detected extensions to work.

I was just pulling copyparty from docker hub at this point. Then I cycled through the heavier images hoping they had the needed dep. I eventually tried djd too because why not.

I ended up figuring what dep was needed and patching the docker file and I've been using that ever since about a month ago.

I also avoided using the script for this as it pulls the latest release from github. So currently I'm building the sfx, placing it where the dockerfile expects and building directly with docker.

It worked so I saw no reason to change to anything else. Plus I like Debian.

Once this dep is in the standard images I intend to stop building my own and pulling from dockerhub again.

9001:
Gotcha -- the images have been built and pushed now, so should be all good. Thanks :>

PR #177, opened 2025-07-01
feat: add various quick searches

ViRb3:
TODO, just putting this up for early testing and feedback.

https://github.com/user-attachments/assets/342dde92-dffe-4d74-a6c5-e4dc3cd94bc2

Screenshot 2025-07-01 at 04 04 45

### DIFF
--- a/copyparty/web/browser.css
+++ b/copyparty/web/browser.css
@@ -1343,7 +1343,7 @@ html.y #widget.open {
 	color: var(--a);
 	text-shadow: 1px 1px 1px var(--op-a-sh);
 	font-size: 1.5em;
-	padding: .25em .4em;
+	padding: 0em .4em;
 	margin: 0;
 }
 #ops a.act {
@@ -1366,6 +1366,13 @@ html.y #ops svg circle {
 #ops {
 	padding: .3em .6em;
 	white-space: nowrap;
+	display: flex;
+	align-items: center;
+}
+
+#opa_filter {
+	margin: 0em .4em;
+	min-width: 50px;
 }
 #noie {
 	color: #b60;
--- a/copyparty/web/browser.js
+++ b/copyparty/web/browser.js
@@ -232,6 +232,10 @@ var Ls = {
 		"ct_idxh": 'show index.html instead of folder listing">htm',
 		"ct_sbars": 'show scrollbars">⟊',
 
+		"qs_off": 'disable quicksearch, hotkeys will work">off',
+		"qs_filter": 'typing without a focused item will hide all items that do not start with this text; disables hotkeys">filter',
+		"qs_jump": 'typing without a focused item will jump to the next item that starts with this text; disables hotkeys">jump',
+		
 		"cut_umod": "if a file already exists on the server, update the server's last-modified timestamp to match your local file (requires write+delete permissions)\">re📅",
 
 		"cut_turbo": "the yolo button, you probably DO NOT want to enable this:$N$Nuse this if you were uploading a huge amount of files and had to restart for some reason, and want to continue the upload ASAP$N$Nthis replaces the hash-check with a simple <em>&quot;does this have the same filesize on the server?&quot;</em> so if the file contents are different it will NOT be uploaded$N$Nyou should turn this off when the upload is done, and then &quot;upload&quot; the same files again to let the client verify them\">turbo",
@@ -855,6 +859,10 @@ var Ls = {
 		"ct_readme": 'vis README.md nedenfor filene">📜 readme',
 		"ct_idxh": 'vis index.html istedenfor fil-liste">htm',
 		"ct_sbars": 'vis rullgardiner / skrollefelt">⟊',
+
+		"qs_off": 'deaktiver hurtigsøk, hurtigtaster vil fungere">av',
+		"qs_filter": 'skriving uten et fokusert element vil skjule alle elementer som ikke starter med denne teksten; deaktiverer hurtigtaster">filter',
+		"qs_jump": 'skriving uten et fokusert element vil hoppe til neste element som starter med denne teksten; deaktiverer hurtigtaster">hopp',
 
 		"cut_umod": 'i tilfelle en fil du laster opp allerede finnes på serveren, så skal serverens tidsstempel oppdateres slik at det stemmer overens med din lokale fil (krever rettighetene write+delete)">re📅',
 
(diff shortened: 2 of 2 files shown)


copyparty/web/browser.css
```diff
@@ -1366,6 +1366,13 @@ html.y #ops svg circle {
 #ops {
 	padding: .3em .6em;
 	white-space: nowrap;
+	display: flex;
```
ViRb3, line 1369: Needed to make the filter input vertically centered

copyparty/web/browser.css
```diff
@@ -1343,7 +1343,7 @@ html.y #widget.open {
 	color: var(--a);
 	text-shadow: 1px 1px 1px var(--op-a-sh);
 	font-size: 1.5em;
-	padding: .25em .4em;
+	padding: 0em .4em;
```
ViRb3, line 1346: Needed to preserve the old look after adding flex to `#ops`

copyparty/web/browser.css
```diff
@@ -1366,6 +1366,13 @@ html.y #ops svg circle {
 #ops {
 	padding: .3em .6em;
 	white-space: nowrap;
+	display: flex;
+	align-items: center;
+}
+
+#opa_filter {
+	margin: 0em .4em;
+	min-width: 50px;
```
ViRb3, line 1375: Default min-width makes this go out of screen on mobile

copyparty/web/browser.js
```diff
@@ -856,6 +860,10 @@ var Ls = {
 		"ct_idxh": 'vis index.html istedenfor fil-liste">htm',
 		"ct_sbars": 'vis rullgardiner / skrollefelt">⟊',
 
+		"qs_off": 'deaktiver hurtigsøk, hurtigtaster vil fungere">av',
```
ViRb3, line 863: AI translated, please verify

copyparty/web/browser.js
```diff
@@ -1480,6 +1488,10 @@ var Ls = {
 		"ct_idxh": '显示 index.html 代替文件夹列表">htm',
 		"ct_sbars": '显示滚动条">⟊',
 
+		"qs_off": '禁用快速搜索，热键将正常工作">关闭',
```
ViRb3, line 1491: AI translated, please verify

copyparty/web/browser.js
```diff
@@ -2156,6 +2286,33 @@ ebi('op_cfg').innerHTML = (
 	'<div><h3>' + L.cl_hiddenc + ' &nbsp;' + (MOBILE ? '<a href="#" id="hcolsh">' + L.cl_hidec + '</a> / ' : '') + '<a href="#" id="hcolsr">' + L.cl_reset + '</a></h3><div id="hcols"></div></div>'
 );
 
+// Bind quicksearch setting to UI
+(function() {
```
ViRb3, line 2292: I could not get `bcfg_bind` to work, feel free to simplify if possible

copyparty/web/browser.js
```diff
@@ -7034,6 +7191,8 @@ var ahotkeys = function (e) {
 	if (aet && aet != 'a' && aet != 'tr' && aet != 'td' && aet != 'div' && aet != 'pre')
 		return;
 
+	if (handle_qsearch_key(e)) return;
```
ViRb3, line 7196: The placement of this is very tricky, but I think this should be right? Please verify

9001:
really cool tech demo!

but I think it would be even more practical if it behaved like firefox's built-in quicksearch;
* global hotkey `/` to move the focus into the filter textbox
* pressing Enter inside the inputbox would first move the focus entirely from the textbox and over to the selected search result, and then click it
* pressing Escape would leave the textbox, and...
  * ...leave the focus at the currently hilighted search result, if any
  * ...or possibly restore the previously focused element otherwise

that way,
* opening a link in mode 2 (basically a clone of firefox's built-in) would just be tapping `/`, then typing a substring, then...
  * cursorkeys up/down iterates through the matches (only one result is hilighted at a time)
  * hitting enter opens the selected link
* and mode 1 would be (possibly, not sure!) something like: tapping `/`, typing the substring, then...
  * enter leaves the textbox and focuses on the first result
  * escape clears the results, restores the original listing, and restores original focus

this would make it possible to keep hotkeys enabled in all of the quicksearch modes, and we could just read the value of the textbox instead of keeping track of the input manually, allowing for full editing / seeing typos.

and then we could redefine mode 0 to entirely hide the Filter-box, since some people may prefer saving the screen space (old iphones are way cramped). And firefox users will probably prefer this too, since they already have quicksearch :>

some general feedback:
* ~~right now it only does files, and not folders -- should probably do those too?~~ EDIT: my bad, thought it was a substring search, but it's instead searching for files starting with the given value.
* right now mode2 doesn't won't work in lazyloaded folders (folders with more than 250 files), and the file you're looking for has not been loaded yet
  * one workaround for this could be to force the whole folder to get rendered into view before initiating the search -- `treectl.lsc.files.length + treectl.lsc.dirs.length` is the current number of items in total, and if that is less than `treectl.nvis` then `treectl.showmore(treectl.lsc.files.length + treectl.lsc.dirs.length)` will trigger the rest to get rendered in
    * it would be best to initiate this logic as soon as the filter textbox gets focus, since loading more entries into the table is not a synchronous action (but it will not delay for more than one rendering cycle, so the delay between gaining focus and the first keystroke is definitely enough)
  * this is one case where our quicksearch might "outperform" the one in firefox, since it doesn't *quite* fit with what copyparty does -- it will successfully scroll to the bottom of page, since the file is currently not rendered in, which triggers the listing to be grown so it now contains the file, but firefox doesn't then scroll down to actually focus on it
* the way it selects files in mode2 is not optimal, since it would interfere with some usecases -- say you've selected a few files already, and now you're looking for another particular file to add, so you try jumping to it... and now it's not obvious what's an actual selection and what's a markup from this new feature :p
  * I think the best approach would be to reuse the markup that's used when you're traversing the files list in cursormode...
    * making the css for `#files tr:focus` also apply to `#files tr.qhit` should do the trick
    * should probably do `#files tr:focus td+td` and the `#files tr:focus:not(.play):...` one too
* it *almost* works in firefox 52.9 and chrome 49, which is the oldest browsers I try to guarantee are still able to run (most of) the javascript, being the final versions released for winXP -- but I think the only parts that *don't* work will automatically disappear with the suggested changes (imitating firefox quicksearch). The only one aside from those is scrollintoview not taking `block` as an option argument
  * I can be in charge of doing the necessary code massage for this, since I'm only aware of a very small handful of active copyparty users who need this (unfortunately with very good reasons)...

as for machine-translated chinese, adding a `//m` suffix to the string is fine -- if someone wants to fix them at some point then there's already a good handful :p

PR #204, opened 2025-07-27
Improve host IP address handling in HttpCli

Tristan-Duin:
Detect if the user provided an IP address or hostname using the ipaddress module. Ensure correct resolution and mapping behavior based on the input type.


9001:
nice, thanks! I'll merge this and put my IPv6 fixes on top of it :>

PR #206, opened 2025-07-27
update systemd stuff for arch

icxes:

Nothing has changed since the last time we talked about this (what can I say, it works for me). If you notice or remember something I guess I guess I can take a stab at it or you could do it yourself. If there's going to be new users I think this version is a bit easier than the one where you have to go and edit systemd service files

Oh, should note that since prior to this people were copying their configs anyway, this change (should) not break anything for existing users, who can (and by default, do) just use their customized service files like before


9001:
thx, good timing -- if we're making this change, then the sooner the better at this point :>

though I think @Morganamilo should make the call -- this is intended to make it easier to specify what user to run copyparty as, without having to edit the systemd files, and also adding a user-service. Is this something we want?

also, heads-up: there will be a new release in less than 12 hours; it would be good to have this packaged for archlinux when that happens.

icxes:
memory (and thinking in general) is a bit hazy due to first being sleep deprived for a few days then sleeping way too much, but as i recall the point of the normal service is so you can run copyparty at whatever location you want in the filesystem (as whatever user you choose), whereas with the user service you're running it as a user program, probably serving on $HOME/copyparty or something

so you would run ```systemctl --user start copyparty``` and it should just run it as your user, and you edit config at $HOME/.config/copyparty (so you don't need root for anything)

while if you wanted to run it as a system service you run ```systemctl start copyparty@foo``` and edit config at /etc/copyparty.d/, you can't do any of this without root

the downside is it still needs the user to know these things (where to edit config/how to get something basic running) which i guess is usually achieved by the program itself having a wizard/the readme being readable/the program having an arch wiki page, but that's a bit out of scope for this

but yeah someone with a slightly better brain for this stuff should look at it before merging at least

9001:
thx for your work btw, good stuff :ok_hand:

icxes:
while testing the user service, I did notice the logs say
```
WARNING: found config files in [/home/fgsfds/.config/copyparty]: copyparty.conf
config files are not expected here, and will NOT be loaded (unless your setup is intentionally hella funky)
```
which would obviously not be true since that's where the user service loads the config from, so that would need to be fixed before any release with these changes. just a cosmetic thing really since everything works fine with the user service, but still

icxes:
~~[87ba7fa](https://github.com/9001/copyparty/pull/206/commits/87ba7fa98babcbdd55952d640200c4465d9cedd7) depends on https://github.com/9001/copyparty/pull/439~~

added in 971360e91434e34095bb9d0a7a9115f7c14fec85

Morganamilo:
User services and templated services both seem useful for copyparty's use case. Though I'm just packaging the software I'm not here to make any calls or review the PR :P

I will say though, I'm not sure why this is specifically in an arch sub directory. A systemd service is generally useful and I don't think it should be tied to any specific distro. It's also useful enough to not be in contrib/ as long as you're willing to maintain it as a first class feature.

icxes:
that's true. the "example" systemd service(s) is/are at contrib/systemd/, so i guess i should probably scrap these and go change stuff there instead. thanks for the heads up

icxes:
closed in favor of #502

PR #211, opened 2025-07-27
docker: remove deprecated version attr

ptweezy:
Removed the deprecated version attribute.

Resolves error "the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion" when building using docker-compose.


9001:
thanks o/

PR #228, opened 2025-07-28
various improvements to the nix files

dtomvan:

Summary:
- Added u2c
- Added partyfuse
- Support adding any (python) packages to the copyparty package
- Cosmetic changes, because everything was unformatted spaghetti...
- Add formatter
- Add extra checks

I haven't looked at the NixOS module as much, but I think I'll make some improvements to that one later.

### DIFF
--- a/contrib/nixos/modules/copyparty.nix
+++ b/contrib/nixos/modules/copyparty.nix
@@ -4,28 +4,31 @@
   lib,
   ...
 }:
-with lib; let
-  mkKeyValue = key: value:
-    if value == true
-    then
+with lib;
+let
+  mkKeyValue =
+    key: value:
+    if value == true then
       # sets with a true boolean value are coerced to just the key name
       key
-    else if value == false
-    then
+    else if value == false then
       # or omitted completely when false
       ""
-    else (generators.mkKeyValueDefault {inherit mkValueString;} ": " key value);
+    else
+      (generators.mkKeyValueDefault { inherit mkValueString; } ": " key value);
 
-  mkAttrsString = value: (generators.toKeyValue {inherit mkKeyValue;} value);
+  mkAttrsString = value: (generators.toKeyValue { inherit mkKeyValue; } value);
 
-  mkValueString = value:
-    if isList value
-    then (concatStringsSep ", " (map mkValueString value))
-    else if isAttrs value
-    then "\n" + (mkAttrsString value)
-    else (generators.mkValueStringDefault {} value);
+  mkValueString =
+    value:
+    if isList value then
+      (concatStringsSep ", " (map mkValueString value))
+    else if isAttrs value then
+      "\n" + (mkAttrsString value)
+    else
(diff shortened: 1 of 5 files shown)


chinponya:
It's good to have an actual nixpkgs contributor look through these.

All changes seem like an improvement to me without breaking anything, so it should be safe to merge this @9001.

It seems like everyone who touched nix in this repo chose to use a different formatter, but since you've included yours in the flake, I'd say we have a winner.

9001:
sweet, was waiting for you to get a chance to peep it :>

thanks a lot for the contribution!

sodiboo:
> It seems like everyone who touched nix in this repo chose to use a different formatter, but since you've included yours in the flake, I'd say we have a winner.

Yeah. It was kinda the wild west for a while. But `nixfmt` 1.0.0 is The Nix Formatter now; and was only recently released. (see: [reformatting nixpkgs](https://github.com/NixOS/nixpkgs/commit/5a0711127cd8b916c3d3128f473388c8c79df0da)).

dtomvan:
Thanks for merging! Now I can use hooks with additional dependencies as well!

Okay, so what I'm still interested in:
- Making the module "nixpkgs ready" to get upstreamed into there
- Packaging this from source, without the dockerfile. I might need a little help with this one, because there's a bunch of dependencies and patches that get pulled which I can't really grok

chinponya:
> * Packaging this from source, without the dockerfile. I might need a little help with this one, because there's a bunch of dependencies and patches that get pulled which I can't really grok

It looks like [someone beat you to it](https://github.com/9001/copyparty/pull/253), but maybe you are interested in collaborating there, in case you find it insufficiently good.
This version uses release source archive, which is actually different from what's on the repo - it contains bundled frontend dependencies (same as the pypi release), sidestepping all of the cursed stuff happening inside dockerfiles.

Strictly speaking that's not a real "built from source" derivation, but it may be a reasonable compromise. Some web dependencies such as `prism` do not have (at least in the past) a way to create a bundle without using their website (though for that [we made our own bundler](https://github.com/9001/copyparty/blob/hovudstraum/scripts/deps-docker/genprism.py) in the previous attempt to package it from source).

dtomvan:
> It looks like [someone beat you to it](https://github.com/9001/copyparty/pull/253), but maybe you are interested in collaborating there, in case you find it insufficiently good.

Yeah, I saw. Very nice! The PR looks fine. I understand it's not entirely "pure" but it's better than "yeah let's just grab the sfx and call it a day" There are a couple packages that are pulled in the dockerfile that _could_ be replaced with preexisting nixpkgs packages or custom derivations, but there's not a lot of juice to squeeze there.

I think there's a much simpler way to improve reproducibility for the web deps and it's [ADD --checksum](https://docs.docker.com/reference/dockerfile/#add---checksum) in the dockerfile. That doesn't require `wget` either, and improves reproducibility not only for the nix package but also for the regular sfx.

PR #242, opened 2025-07-28
Disable the "use the real password" button on the connect page when there's no accounts

toast003:
If you go to the connect page on a copyparty instance that has no accounts, the button to use a real password still shows up, while placeholder it changes doesn't.

Using it then just makes the UI explode so the button shouldn't show up


9001:
wait... that wiped your commit details?? dangit

9001:
ahh sorry, well thanks for the fix, I'll mention you in the releasenotes at least :>

toast003:
Thanks! <3

PR #246, opened 2025-07-28
cbz thumbnails: sort alphabetically

Scotsguy:
Comic readers will sort alphabetically, but that isn't always the order in which the files are stored in the zip, so sort it before picking a thumbnail.


9001:
awesome thanks :grin:

PR #284, opened 2025-07-29
Add `uv tool run copyparty` to readme

kleutzinger:
I've confirmed that this command seems to work on both mac and linux.

We could go with `uvx copyparty` or `uv tool run copyparty` as they are [aliases of each other](https://docs.astral.sh/uv/guides/tools/). I'd favor `uv tool run` for clarity, but it's no big difference.


chocolateimage:
Can also confirm that it also works on Windows with a fresh Windows 10 VM with both `uvx copyparty` and `uv tool run copyparty`

9001:
thanks :>

PR #296, opened 2025-07-30
nix: make usage in non-flake setups easier

Sorixelle:
using the nix expressions in setups where flakes aren't in use was possible, but a little unwieldy (you had to reconstruct the overlay defined in the flake to get copyparty in your nixpkgs set). this pr moves the overlay into it's own dedicated file, allowing it to be used directly without requiring something like [flake-compat](https://git.lix.systems/lix-project/flake-compat). there's also some extra documentation, illustrating how to use the module in a non-flake nixos setup.

~~i've also added a commit that removes a (seemingly) unnecessary evaluation of nixpkgs, to avoid contributing to the ["1000 instances of nixpkgs"](https://zimbatm.com/notes/1000-instances-of-nixpkgs) issue. this one might be a little controversial, so i've added it as a suggestion - if you're not interested in this change, i'll happily drop the commit. it's not super necessary for anything, just a potential optimisation i spotted.~~

this pr complies with the dco; https://developercertificate.org/

### DIFF
--- a/README.md
+++ b/README.md
@@ -2344,7 +2344,7 @@ some recommended dependencies are enabled by default; [override the package](htt
 
 ## nixos module
 
-for this setup, you will need a [flake-enabled](https://nixos.wiki/wiki/Flakes) installation of NixOS.
+for [flake-enabled](https://nixos.wiki/wiki/Flakes) installations of NixOS:
 
 ```nix
 {
@@ -2368,6 +2368,33 @@ for this setup, you will need a [flake-enabled](https://nixos.wiki/wiki/Flakes) 
       ];
     };
   };
+}
+```
+
+if you don't use a flake in your configuration, you can use other dependency management tools like [npins](https://github.com/andir/npins), [niv](https://github.com/nmattia/niv), or even plain [`fetchTarball`](https://nix.dev/manual/nix/stable/language/builtins#builtins-fetchTarball), like so:
+
+```nix
+{ pkgs, ... }:
+
+let
+  # npins example, adjust for your setup. copyparty should be a path to the downloaded repo
+  # for niv, just replace the npins folder import with the sources.nix file
+  copyparty = (import ./npins).copyparty;
+
+  # or with fetchTarball:
+  copyparty = fetchTarball "https://github.com/9001/copyparty/archive/hovudstraum.tar.gz";
+in
+
+{
+  # load the copyparty NixOS module
+  imports = [ "${copyparty}/contrib/nixos/modules/copyparty.nix" ];
+
+  # add the copyparty overlay to expose the package to the module
+  nixpkgs.overlays = [ "${copyparty}/contrib/package/nix/overlay.nix" ];
+  # (optional) install the package globally
+  environment.systemPackages = [ pkgs.copyparty ];
+  # configure the copyparty module
+  services.copyparty.enable = true;
 }
 ```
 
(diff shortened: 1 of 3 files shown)


chinponya:
LGTM. I don't see a reason why it should be dropped.

It might just conflict a little with https://github.com/9001/copyparty/pull/253 depending on which one gets merged first.

Sorixelle:
the main issue with that `appliedOverlay` trick is that it doesn't work when the overlay expects `final` to be the actual final nixpkgs set (ie. it uses `final` to reference things the overlay itself defines, or things from other required overlays). that's not happening here though, so i figured it would be fine, but if the overlay does get a little more complex and relies on `final` actually being what it should, you'd probably have to drop that trick and just eat the extra instantiation.

9001:
I merged the other nix-related PR just now (didn't realize it was ready to merge until yesterday, sorry) -- is this one still good to go, or will something need to be adjusted?

Sorixelle:
i'll clean up the merge conflicts later today - should be good to merge after that.

Sorixelle:
sorry about the delay! rebase is all done, and ready to merge. i ended up dropping the commit removing the flake optimization, since the new version of the overlay does actually rely on the semantics of `final` now, so the trick in that commit won't work.

9001:
thanks! :>

atroxiv:
Hello, I think something's gone wrong with this. 
I'm attempting to install copyparty on nix using npins. I've pinned it with npins and it's up to date, but the overlay.nix file does not appear to exist in the nix store whatsoever. Naturally, this is causing my attempt to copy the example in the README to fail.

It should be noted that I am a complete noob at nix OS, having just installed it yesterday, so I may just be completely wrong.

image
image
image


I could just use the package version (nix profile install), but I think having a separate conf file is considered heresy on nix or something.

What do?

@9001 @Sorixelle

9001:
sorry but I can't help, I don't know anything about nixos -- hope it works out tho

Sorixelle:
npins pins to the latest tagged release by default, and i don't think this pr has landed in a release yet. you'll want to either wait for a new release, or do `npins add github 9001 copyparty -b hovudstraum` for now to follow the main branch.

atroxiv:
> npins pins to the latest tagged release by default, and i don't think this pr has landed in a release yet. you'll want to either wait for a new release, or do `npins add github 9001 copyparty -b hovudstraum` for now to follow the main branch.

I've done that and attempted to rebuild. The `overlay.nix` file is now present in the nix store (and identical to the file in the github), but it still shows up with the same error.

Sorixelle:
oh crap, i didn't read that error close enough. my documentation is wrong - you'll need `(import "${copyparty}/contrib/package/nix/overlay.nix")` in there instead of just the plain string. sorry about that - i'll pr a fix for that asap!

atroxiv:
> oh crap, i didn't read that error close enough. my documentation is wrong - you'll need `(import "${copyparty}/contrib/package/nix/overlay.nix")` in there instead of just the plain string. sorry about that - i'll pr a fix for that asap!

Glad I'm the only one affected, then! Thanks for the hard work, I can see this is a **bit** technical, so no need to apologize haha.

atroxiv:
> > oh crap, i didn't read that error close enough. my documentation is wrong - you'll need `(import "${copyparty}/contrib/package/nix/overlay.nix")` in there instead of just the plain string. sorry about that - i'll pr a fix for that asap!
> 
> Glad I'm the only one affected, then! Thanks for the hard work, I can see this is a **bit** technical, so no need to apologize haha.

It builds correctly with that fix btw! Finally get to use my new file server lol.

PR #305, opened 2025-07-30
Update README.md Quickstart

ChristophorusReyhan:
quickstart for users who want to use config file instead

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,8 @@ made in Norway 🇳🇴
 
 ## quickstart
 
+### 1-Line Install (Instant Run)
+
 just run **[copyparty-sfx.py](https://github.com/9001/copyparty/releases/latest/download/copyparty-sfx.py)** -- that's it! 🎉
 
 * or install through [pypi](https://pypi.org/project/copyparty/): `python3 -m pip install --user -U copyparty`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -159,7 +161,60 @@ just run **[copyparty-sfx.py](https://github.com/9001/copyparty/releases/latest/
...
+   ./dropbox
+   accs:
+     w: bob
+ ```
+ Start with:
+ ```bash
+ python copyparty-sfx.py -c config.conf
+ ```
+
+ **4. Install Extras (Optional)**
```
bytequill, line 203: consider rewording Extras into `extra/optional dependencies`

README.md
```diff
@@ -159,7 +161,60 @@ just run **[copyparty-sfx.py](https://github.com/9001/copyparty/releases/latest/
 * or if you prefer to [use docker](./scripts/docker/) 🐋 you can do that too
   * docker has all deps built-in, so skip this step:
```
bytequill, line 162: The config-file install header breaks the flow of this line into the `#Thumbnail` instructions. Consider rewording and/or adding a link to the section?

eg.
```suggestion
  * docker has all deps built-in, so, skip the [Thumbnails](#thumbnails) step
```

README.md
```diff
@@ -159,7 +161,60 @@ just run **[copyparty-sfx.py](https://github.com/9001/copyparty/releases/latest/
...
+ python copyparty-sfx.py -c config.conf
+ ```
+
+ **4. Install Extras (Optional)**
+ 
+ See [Thumbnails Section](#thumbnails) for more
+ 
+ Ubuntu/Debian: `sudo apt install python3-pil ffmpeg`
+ 
+ Windows: `pip install --user Pillow` + add FFmpeg to PATH
```
bytequill, line 209: 1. add the `--no-install-recommends` to the `apt` example (as seen on [L220](https://github.com/9001/copyparty/blob/559f72770c4a1c5e733733983b19cc72f3e80084/README.md?plain=1#L220)) 
2. While I understand including install instructions for most used OSes it would probably be better to *just* link to the section because this way you include all the additional Windows information, basic documentation on media tags

README.md
```diff
@@ -159,7 +161,60 @@ just run **[copyparty-sfx.py](https://github.com/9001/copyparty/releases/latest/
...
+
+ **5. Expose Publicly (Optional)**
```
bytequill, line 211: There are a **lot** of ways to expose a web service on the open internet. Cloudflare tunnel is a good way to securely expose ad-hoc but I'm not sure we want to specifically feature that in a quickstart

bytequill (commented):
Just a bit of minor feedback (mostly just pet-peeves) from a *random* person, no idea why I decided to randomly review a doc PR

fantasmaancargel2009-afk (changes requested):
tth: virus&+??‽

PR #310, opened 2025-07-30
translating to Spanish

herruzo99:
PR to add Spanish language to the project. For any native Spanish speaker free to update note improvements / changes to the translation.

The translation was first made with an LLM (no code generation, just the translation) to translate the bulk of the lines, and then checked and fixed line by line (native Spanish speaker).

~~I have seen that the control plane cannot (or i did not find how to) be currently translated neither the headers of the grid view of the files, so those ones is kept in English.~~ EDIT: it was in `copyparty/web/splash.js`


copyparty/web/browser.js
```diff
@@ -1884,9 +1884,634 @@ var Ls = {
...
+
+		"cl_opts": "opciones",
+		"cl_themes": "tema",
+		"cl_langs": "idioma",
+		"cl_ziptype": "descarga de carpeta",
+		"cl_uopts": "opciones up2k",
+		"cl_favico": "favicon",
+		"cl_bigdir": "directorios grandes",
+		"cl_hsort": "#ordenar",
+		"cl_keytype": "notación de teclas",
```
toast003: Imo this should be `Notación musical`, cause the current translation used the wrong translation of key.
I'm not entirely sure this is the correct translation either since I don't know the right terminology in Spanish, but it's def more correct than the current translation

copyparty/web/browser.js
```diff
@@ -1884,9 +1884,634 @@ var Ls = {
...
+		"tt_detree": "mostrar breadcrumbs$NAtajo: B",
+		"tt_visdir": "desplazarse a la carpeta seleccionada",
+		"tt_ftree": "alternar árbol de carpetas / archivos de texto$NAtajo: V",
+		"tt_pdock": "mostrar carpetas de niveles superiores en un panel acoplado en la parte superior",
+		"tt_dynt": "crecimiento automático a medida que el árbol se expande",
+		"tt_wrap": "ajuste de línea",
+		"tt_hover": "revelar líneas que se desbordan al pasar el ratón$N( rompe el desplazamiento a menos que el $N&nbsp; cursor esté en el margen izquierdo )",
+
+		"ml_pmode": "al final de la carpeta...",
+		"ml_btns": "cmds",
```
toast003: I would translate this to `comandos` or `acciones`, we have the space since the translation for the button is also longer

copyparty/web/browser.js
```diff
@@ -1884,9 +1884,634 @@ var Ls = {
...
+		"ml_tcode": "transcodificar",
+		"ml_tcode2": "transcodificar a",
+		"ml_tint": "tinte",
+		"ml_eq": "ecualizador de audio",
+		"ml_drc": "compresor de rango dinámico",
+
+		"mt_loop": 'poner en bucle/repetir una canción">🔁',
+		"mt_one": 'parar después de una canción">1️⃣',
+		"mt_shuf": 'reproducir aleatoriamente las canciones en cada carpeta">🔀',
+		"mt_aplay": 'reproducción automática si hay un ID de canción en el enlace en el que hiciste clic para acceder al servidor$N$Ndesactivar esto también evitará que la URL de la página se actualice con IDs de canción al reproducir música, para prevenir la reproducción automática si se pierden estos ajustes pero la URL permanece">a▶',
```
toast003: `reproducir automaticamente` instead of `reproducción automática`

copyparty/web/browser.js
herruzo99: Yeah, i looked into the context of where the string is used and i would think that is a good translation, i am not fluent in music terminology nevertheless that is a better translation. Thanks!

toast003: No problem ^^

9001:
just in case you're any better at git than I am, would you be able to rebase this? otherwise I'll keep trying and maybe eventually figure it out :grin:  just keep your translation right below german so they're alphabetic if possible

DaitoIsBlank:
Also please change "esp" to "spa" to comply the ISO 639-2 standard 🫡

herruzo99:
@9001 I think i  have just rebased it correctly

copyparty/web/browser.js
```diff
@@ -4390,7 +4390,7 @@ var Ls = {
 	}
 };
 
-var LANGS = ["eng", "nor", "chi", "deu", "rus", "ukr", "spa"];
+var LANGS = ["eng", "nor", "chi", "deu", "spa", "rus", "ukr"];
```
icxes, line 4393: R comes before S in the alphabet

copyparty/web/splash.js
```diff
@@ -94,6 +94,7 @@ var Ls = {
 		"af1": "显示最近上传的文件", //m
 		"ag1": "查看已知 IdP 用户", //m
 	},
+
```
icxes, line 97: are these necessary? if anything, i'd remove the newlines between each language block

9001:
thanks! gonna hurry and merge this because there's lots of upcoming stuff in browser.js

PR #317, opened 2025-07-30
nix: proposed setup to hide `nixpkgs.overlays` from users

takagemacoed:
This may conflicts with the PR https://github.com/9001/copyparty/pull/296 (for non-flake setup)

I will do a force push for this PR to make it also compatible for non-flake setup after that PR merge

### DIFF
--- a/README.md
+++ b/README.md
@@ -2278,8 +2278,6 @@ for this setup, you will need a [flake-enabled](https://nixos.wiki/wiki/Flakes) 
         # load the copyparty NixOS module
         copyparty.nixosModules.default
         ({ pkgs, ... }: {
-          # add the copyparty overlay to expose the package to the module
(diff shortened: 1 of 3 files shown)


toast003:
Imo we should probably add a note (either on the README or the changelog) telling people to not import the overlay manually, since importing the overlay twice [can break things](https://github.com/Jovian-Experiments/Jovian-NixOS/pull/487#issue-2885098647l)

takagemacoed:
> Imo we should probably add a note (either on the README or the changelog) telling people to not import the overlay manually, since importing the overlay twice [can break things](https://github.com/Jovian-Experiments/Jovian-NixOS/pull/487#issue-2885098647l)

Nice, I did not think about that. I think it is better not to touch it for now.

PR #321, opened 2025-07-30
Translating to Russian

A1Asriel:
This pull request aims to translate every localizable string to Russian. No LLMs or machine translation engines were used for this contribution (Edit: Had to use Google Translate for a single string (`ag1`) which I couldn't find translated to English in the project). A native Russian speaker performed the translation.

**Note:** a few style decisions were made to conform to the grammar of the target language. For instance, the excessive use of hyphens was eliminated, and periods were added to the end of some truncated words. This isn't a completely necessary step, but it might make the text easier to read, especially for those who aren't tech savvy.
Most of the jokes are either adapted to native language, or left as-is, albeit translated to Russian.

While it is unclear whether the not-yet-translated strings will be compact enough to avoid stretching the elements too much while remaining understandable to the user, testing of the current state shows that the UI flow does not change drastically.
Edit: couldn't really test most of the errors, but all normally-accessible text strings look fine.

---

> This PR complies with the DCO; https://developercertificate.org/


9001:
I see someone made another pr over at #363 -- you were first though so I'd like to say that "you win" :-) but would be worth a look to see if there's anything to grab from there?

A1Asriel:
> I see someone made another pr over at #363 -- you were first though so I'd like to say that "you win" :-) but would be worth a look to see if there's anything to grab from there?

I just finished my translation so I'll flag it as ready for review, but sure, I'll take a look at that version as well!

A1Asriel:
Okay, my version is far more verbose than the newer one, yet I see they've removed some of the jokes or explanations (like for the compressor), and didn't translate any of the button labels. While I see the appeal of using more tech slang (which sounds mostly the same in Russian, just transliterated to Cyrillic), I believe they have to be translated for the average user who might not be able to read English at all. Which is why I kept usage of English to minimum, limited only to product names or very specific terms.

Some of the wordings might sound better in their version, and I have nothing against it. Our language is quite flexible in this aspect ^^ Many decisions here can be argued over, like the usage of "ты" or singular "вы" (less casual, more neutral/respectful)

So I'd say push my translation to master and then let anyone edit it, if needed :)

PR #351, opened 2025-07-31
Nix: remove space from list separator

toast003:
The extra spaces made copyparty freak out if you tried listening to an ip address and a unix socket at the same time.


Daimanta (approved):

chinponya:
Yep, that was actually just a bug.

IIRC I only ever tried it when listing users for a given set of permissions, which allowed spaces, but removing them doesn't seem hurt anything.

chinponya (approved):

PR #374, opened 2025-07-31
Add info about pipx and uvx

bb:


bb:
Just noticed that there's also https://github.com/9001/copyparty/pull/284

9001:
I'll go with #284 but thanks for the headsup re pipx, will add that too :+1:

PR #377, opened 2025-07-31
chore: updated authelia docker-compose.yml

xFuture603:
This pull request adds some improvements for the current example `docker-compose.yml`, configured with Authelia.

It adds:
- Valkey instead of Redis (even though this project does not resell Redis, Valkey is backed by the Linux Foundation and was essentially replaced via drop-in)
- adds password to the valkey configuration
- included a docker socket proxy to not bind the docker socket directly to Traefik (improves security if an attacker hijacks the Traefik instance via an exploit, for example)
- pin images to the currently latest available versions and their SHAs
- fixed some Authelia configuration warnings
- updated Traefik from v2 to v3 (currently latest available version)
- updated README.md


xFuture603:
Hey @9001. I just wondered if there is anything I can help with to get this pull request considered? Thanks in advance!

9001:
hey o/

just some quick questions,

* why is the empty acme.json there?
* not sure if it's a good idea to pin the copyparty version

for example, the version which is pinned in this changeset has a denial-of-service vulnerability. Having the version pinned implies (to me) that it should be increased every time a new version is released, but, at the same time, I left all the other images pinned to indicate which exact combination has been tested to work. Honestly not sure what the right call is here... What do you think?

xFuture603:
Thanks for pointing that out! I removed acme.json. It was just left over from testing. Otherwise, I pinned the image version of copyparty back to “latest.” I pinned the remaining images to one version because the readme said: “_Pin to image hashes (or go all in on updates and add `watchtower`)_”. 
This would allow the images to be automatically updated in the repository via an auto-update mechanism such as Renovate after testing.

Is that okay with you? Let me know!

9001:
looks good! thanks for helping out with this :>

PR #385, opened 2025-08-01
add Debian packaging via MPR

Beethoven-n:
this took most of my day but i'm happy with the result.
you should even be able to distribute the resulting package on debian-based distros.

the MPR is a debian equivalent to the arch AUR. instead of making arch packages, it makes debian .deb packages.
there's even a repository for them! https://mpr.makedeb.org/

there are two optdepends that don't have equivalents as of debian 12, those being `python-pyvips` and `libkeyfinder`. you'll miss out on higher quality thumbnails and musical key detection, but those packages just aren't available.

installation is simple:
* install makedeb `bash -ci "$(wget -qO - 'https://shlink.makedeb.org/install')"`
* clone the repo, copy `contrib/package/makedeb-mpr/PKGBUILD` somewhere (or get the raw)
* `makedeb -s` to build the package
* `sudo apt install ./copyparty_$version_$architecture.deb` to install
* `sudo apt purge copyparty` to uninstall

this PR complies with the DCO; https://developercertificate.org/

### DIFF
--- a/contrib/package/makedeb-mpr/PKGBUILD
+++ b/contrib/package/makedeb-mpr/PKGBUILD
@@ -1,0 +1,46 @@
+# Maintainer: Morgan Adamiec <morganamilo@archlinux.org>
+# Contributor: icxes <dev.null@need.moe>
+# makedeb PKGBUILD script writer: Beethoven <beethovenisadog@protonmail.com>
+
+
+pkgname=copyparty
+pkgver=1.18.8
+pkgrel=1
+pkgdesc="File server with accelerated resumable uploads, dedup, WebDAV, FTP, TFTP, zeroconf, media indexer, thumbnails++"
+arch=("any")
+url="https://github.com/9001/${pkgname}"
+license=('MIT')
+depends=("bash" "python3" "lsof" "python3-jinja2")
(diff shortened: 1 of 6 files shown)


Beethoven-n:
you could actually distribute the resulting .deb file in releases but i didn't put anything in directly to do that. maybe something for the `scripts` folder

icxes:
I probably should've mentioned it earlier, but there are plans to change how the services work to make the setup a little simpler: see #206. It's just been held back a little since I'd like to get the maintainer of the Arch package to chime in on the changes first.

Beethoven-n:
oh, didn't know that. understood!

9001:
looks good, though I wonder if it would be a good idea to remove the original maintainer/contributor names from the makedeb-mpr/PKGBUILD since this isn't something they are immediately responsible for, could instead leave a comment that it was autogenerated from the archlinux PKGBUILD and leave the breadcrumbs that way?

Beethoven-n:
i could do that
it wasn't autogenerated though, i don't know how to do that so i just manually resolved the packages myself.
does that confer any extra responsibilities?
i kinda just wanted to make the pkgbuild and leave it at that

9001:
thanks, looks good!

PR #406, opened 2025-08-01
[Fix] Changed "ack" to "Back" in copyparty/httpcli.py

JanluOfficial:
# About this commit
This commit fixes the hyperlinks defined in lines 2943, 2972 and 2986 to say "Back" instead of "ack", as they did before. Even though this wasn't a huge deal breaker, it also didn't take long to fix and improves the overall experience in the Web UI.

# AI usage and DCO compliancy
This commit **does not involve any use of AI/LLMs** (GitHub Copilot, ChatGPT, Gemini) and therefore **complies with the** [**contribution guidelines**](https://github.com/9001/copyparty/blob/ad23b253dc033ef19eace3816bc6976b1b38090b/CONTRIBUTING.md#do-not-use-ai--lmm-when-writing-code) of this repository.


JanluOfficial:
I did not mean for [Fixed a critical syntax mistake in browser.js](https://github.com/9001/copyparty/pull/406/commits/5c3d6e49a540d8b00809dd111939f2209b4c458e) to be in this commit, however, I will explain the changes really quickly.

On line 3077 in copyparty/web/browser.js, there was a string with quotations. These quotations did not have backslashes in front of them though, causing this error.
image

I quickly added the missing backslashes, which got everything up and running again.

### Before

> `		"u_enoow": "ylikirjoitus ei toimi täällä; tarvitaan "Delete"-oikeus",`

### After

> `		"u_enoow": "ylikirjoitus ei toimi täällä; tarvitaan \"Delete\"-oikeus",`

In case this additional commit invalidates this PR, I will make separate pull requests for each. Sorry for any inconveniences this may cause.

icxes:
Alright, about the error, that's embarrassing. That's completely my fault. Thanks for catching that before it reached anyone! As for this PR, I don't remember it ever not being "ack" (which I assumed was short for "acknowledged"). 

For example, here's the handle_login function from 6 years ago:
https://github.com/9001/copyparty/blob/v0.2.3/copyparty/httpcli.py#L355

JanluOfficial:
> Alright, about the error, that's embarrassing. That's completely my fault. Thanks for catching that before it reached anyone! As for this PR, I don't remember it ever not being "ack" (which I assumed was short for "acknowledged").
> 
> For example, here's the handle_login function from 6 years ago: https://github.com/9001/copyparty/blob/v0.2.3/copyparty/httpcli.py#L355

I guess "ack" being short for "acknowledged" would make a lot more sense, though all that button does is return you to the main interface which is why I thought it was supposed to be "back". Changing it to "acknowledged" would still improve the experience though so I'm open to doing that. I'd say another opinion on this would be great though.

JanluOfficial:
Now that I think about it, changing "ack" to "Acknowledge" doesn't really make enough of a difference. I'll submit the fix for copyparty/web/browser.js as a completely separate PR but close this one since I think the rest is unnecessary. 

Edit: There's already a fix so... yea

PR #426, opened 2025-08-02
Re-add Dutch translation

DeStilleGast:
as a native Dutch speaker, reader and writer but still used some translation tools, Dutch is hard 😅
I did change some lines to be more fitting such as `folder download` to `download folder as` (but translated) as it would be really wierd.

I did try to check all pages and none of them seems broken.


9001:
thanks, I'll take a look tongiht :>

9001:
awesome, that's much better :grin: 

i think the only thing missing is the translation for ct_utc and this should be good to go!

DeStilleGast:
👀 it wasn't there before, will add that line

9001:
thanks!

PR #439, opened 2025-08-02
add PRTY_CONFIG environment variable for loading the config file

icxes:

not sure if this is the best or even a smart way of doing this but it is what it is


9001:
awesome idea, thx -- will update the readme when it gets closer to the next ver,

Screenshot_2025-08-02_23-16-16

PR #456, opened 2025-08-03
add the ability to transcode to wav and flac formats

missaustraliana:
### _tl;dr this pull request adds the ability to transcode to wav and flac formats for alac support_

hey!

this pull request opens the gate for copyparty to be more versatile and flexible in its codecs. by adding the option to transcode to flac and wav codecs, files with apple's alac codec can now be played directly from the browser, without any sort of quality loss.

alac has not been adopted in modern browsers today. while it's nearly identical to flac, it's support has been neglected by every browser except apple's own safari. because alac, flac and wav are so similar, we can transcode super easily between each other with no quality loss. 

transcoding to wav is crucial since it's support has been around for [ages](https://caniuse.com/wav). giving even the oldest browsers and systems, the ability to play back even the highest sample rates. 
transcoding to wav can also be extremely handy for other codecs like mp3! if a user were to play an opus file and the browser didnt support it, it would have to be transcoded to mp3. thats 2 sets of audio compression resulting in lower fidelity. if the user were to click transcode to wav, they would be able to play back the opus files contents with no degraded quality.

i wish that that you merge this pr to help with audio fidelity in your software.

many thanks

_This PR complies with the DCO; https://developercertificate.org/_


9001:
nice! you even tiptoed around the webp/wav collision :-)

I'll merge this, but I'll make both flac and wav default-disabled to prevent pranksters from filling the server-HDD with transcodes, since they're so much larger than opus/mp3 ones.

thanks for the contribution :grin:

PR #458, opened 2025-08-03
Translating to Italian

AOTREVAI:
Added italian translation
translation was made from the english version
Claude LLM was used for a rough initial translation, but each phrase was double checked (by native Italian speaker) to ensure term consistency, coherence and comprehention.
tested locally for any format or borders problem, seems ok.


9001:
thanks! think you could move the translation so it's alphabetical with the others? Norwegian and English are only kept first because it's the two languages I can proofread :>

I'll take a closer look later tonight

AOTREVAI:
done

9001:
sweet, looks good - only question i have is if the trailing apostrophe is supposed to be there in `un_nou` , aside from that i think this is good to go :>

PR #468, opened 2025-08-04
Greek translation

chamdim:


9001:
Thanks!

Sorry for asking, but can you please confirm if you speak greek natively, and that you've proofread the translation for accuracy?

Could you please move your translation up between russian and spanish, so it's alphabetic?

and there are two lines missing (they were added very recently), right below `mt_c2mp3` there is now also `mt_c2flac` and `mt_c2wav`

Also, if you would like to also translate the controlpanel then there is another file, `splash.js` for that -- here is a template for the strings,  inside tl_cpanel : https://github.com/9001/copyparty/blob/hovudstraum/scripts/tl.js

chamdim:
Thanks for your reply. 

I’m Greek and I’ve typed and reviewed the translation line by line. (it took me roughly 8 hours because I’m not so familiar with JS syntax and didn’t want to mess things up with the values and keys. I would have preferred using something like Weblate like others, but we work with what we’ve got  ; )) 

I’m not sure I understand what you mean by the alphabetical order you mentioned. Do you mean the "g" from "grc" or something else?

I’ll take care of the rest of your requests tomorrow.

Keep up the awesome work on this project.

9001:
sweet!! thanks :>

don't worry about the alphabetic part, I think I can combine it with some cleanup as well, so it's fine to leave it where it is.

good work and looking forward to the final bit :>

chamdim:
Keep in mind that I’ll be revisiting the translation frequently for fine-tuning, since many of the strings I translated didn’t actually show up in the web UI.

*off-topic but for your information, I’ve deployed Copyparty in a Docker container on my server, and I’m managing it from my phone using MiXplorer and WebDAV, and from my PC and laptop via an rclone mount with a systemd service that runs on startup.

PR #471, opened 2025-08-04
Czech translation

kubakubakuba:
_This PR complies with the DCO; https://developercertificate.org/_  

Added a Czech translation, I will proofread it again more times, because mistakes can definitely appear.

I am not certain as how copyparty handles declensions, but in czech, its supposed to be:
1 _sekunda_, 2 - 4 _sekundy_, 5+ _sekund_. The same with hours and days.

So I added:
```json
"ht_s1": "sekunda",
"ht_s2": "sekundy",
"ht_s5": "sekund",
```

I would assume other languages have similar things, so if you would like to implement something similar, there is this option.


9001:
Thanks!

Looks good -- I'll see if I can make it smart enough to use s5/m5/h5 when that's appropriate, but worst case it'll use 2 for all values 2 or higher.

Also, if you would like to also translate the controlpanel then there is another file, splash.js for that -- here is a template for the strings, inside tl_cpanel : https://github.com/9001/copyparty/blob/hovudstraum/scripts/tl.js

aside from that, looks great :+1:

kubakubakuba:
Sure, will look into that!

9001:
oh and one more thing -- if you could move the strings themselves up to between chinese and german then that would be great :>

kubakubakuba:
Should be updated now :)

9001:
yep, looks good!

I can merge this now and make a beta-build of the sfx.py if you want to test it out that way, or would you prefer something else?

I'm not sure when the next version will be, current guess is saturday.

kubakubakuba:
Its ok, I built the sfx locally, it looked like that it did work. Will be looking for the inevitable mistakes for now ;)
Thanks for the great app btw!

PR #515, opened 2025-08-07
add support for systemd socket activation (fd passing)

mati1210:
if running copyparty on a regular computer where it's only going to be occasionally used, it might be useful to only start it if someone tries connecting to it

## Testing

set up a server with some socket-activated interfaces and some regular ones
```bash
export PYTHONPATH=$PWD/copyparty:$PYTHONPATH
systemd-socket-activate --listen 22720  --listen /tmp/sd-copyparty.sock \
  -- python -m copyparty \
      -i fd:3,fd:4,unix:/tmp/copyparty.sock,:: -p 22721
```
then on another terminal try connecting to it
```bash
curl 127.1:22721 -fsS -o /dev/null # this should fail, because the server isn't active yet
curl 127.1:22720 -fsS -o /dev/null # connecting to this should activate the server
curl --unix-socket /tmp/copyparty.sock http://./ -fsS -o /dev/null # the server is already active, so all interfaces should be working
curl --unix-socket /tmp/sd-copyparty.sock http://./ -fsS -o /dev/null 

```

<sub>This PR complies with the DCO; https://developercertificate.org/</sub>

### DIFF
--- a/copyparty/__main__.py
+++ b/copyparty/__main__.py
@@ -564,7 +564,7 @@ def get_sects():
 
             \033[32m-i fd:\033[33m3\033[0m uses the socket passed to copyparty on file descriptor 3
 
-            \033[33m-p\033[0m (tcp ports) is ignored for unix sockets
+            \033[33m-p\033[0m (tcp ports) is ignored for non ip-addresses
             """
             ),
         ],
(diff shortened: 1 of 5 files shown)


9001:
this is cool!! thx :grin:

PR #525, opened 2025-08-08
Add Ukrainian translations for the control panel 🇺🇦

MrMebelMan:
@9001 wassup sjef, adding the forgotten splash.js translations as you requested. Tested it on the login page.

I've also removed the spaces from ul_btnu UA text because I've noticed that it's overflowing cause it's too long (didn't look nice).

Skål!

---


9001:
thanks!! great work as always 😁🎉

PR #537, opened 2025-08-08
legg åt nynorskoversetjing, og fikser et par typos på bokmål også :)

chinatsu:
Oversetjinga er gjort for hand, og eg har kanskje gått litt fort i svingane, så ver oppmerksam på at det kan ha oppstått nokre feil her og der :saluting_face: 

(TL note: I've added [nynorsk](https://en.wikipedia.org/wiki/Nynorsk) as a language)

For å vise at bidraget ditt er kompatibelt med MIT-lisensen, ver venleg å inkludér denne teksten ein plass i denne PR-beskrivelsen.
Denne PR'en overheld DCO; https://developercertificate.org/


chinatsu:
hmm, looks like i'm running into some issues now that i've forked and branched. it was working fine before that :thinking: 

```
/.cpr/browser.js @10903:5
TypeError: can't access property "onclick", ebi(...) is null

UA: Mozilla/5.0 (X11; Linux x86_64; rv:141.0) Gecko/20100101 Firefox/141.0 
```

i'll investigate a bit

9001:
ah one thing that might help (if the output is understandable) is `./scripts/tlcheck.sh eng nor copyparty/web/browser.js`

and don't forget splash.js :D

chinatsu:
> ah one thing that might help (if the output is understandable) is ./scripts/tlcheck.sh eng nor copyparty/web/browser.js

tlcheck reported two diffs, but it's because i've used apostrophes a tiny bit more. 
in my latest commit, i changed `URLen` to `URL'en` in the `"nor"` section as it had been used other places in that set, so there's only one diff according tlcheck:
i chose to use `equalizeren` instead of `equalizer'n`, because it feels a little more natural for nynorsk.

splash.js has been sorted out :saluting_face: 

---

do note, i think i've gone overboard on the silliness using "åt" pretty much everywhere in place of "til" (except two places, where it felt extremely weird to use åt). it's not exacly a natural way to use that word. if it's _too_ silly, let me know and i'll fix the most egregious cases.

9001:
nah this is good, in fact I think you saved us -- the other guy was thinking of taking it way further xD

thanks o/

PR #551, opened 2025-08-09
Add Swedish translation

Bevinsky:
Here's a WIP of a Swedish translation.

I'm not all too comfortable with technical Swedish translation since I only use systems in English, so someone else should probably give this a look-over.


copyparty/web/browser.js
```diff
@@ -8185,6 +8185,635 @@ var Ls = {
...
+				["Blanksteg", "växla val av fil"],
+				["↑/↓", "flytta filvalsmarkör"],
+				["ctrl ↑/↓", "flytta markör och vy"],
+				["⇧ ↑/↓", "välj föregående/nästa fil"],
+				["ctrl-A", "välj alla filer / mappar"],
+			], [
+				"navigation",
+				["B", "växla mellan brödsmulor / trädvy"],
+				["I/K", "föregående/nästa mapp"],
+				["M", "hoppa till överordnad mapp (eller kollapsa närvarande)"],
```
Bevinsky, line 8866: "Överordnad mapp" sounds incredibly formal to me, but it's apparently what it's called.

copyparty/web/splash.js
```diff
@@ -469,6 +469,47 @@ var Ls = {
...
+		"ta1": "fyll i ditt nya lösenord",
+		"ta2": "upprepa det nya lösenordet:",
+		"ta3": "det blev fel; vänligen försök igen",
+		"aa1": "inkommande filer:",
+		"ab1": "avaktivera no304",
+		"ac1": "aktivera no304",
+		"ad1": "detta stänger av all cachning; prova detta om k304 inte räckte till. Detta kommer att slösa enorma mängder nätverkstrafik!",
+		"ae1": "aktiva nedladdningar:",
+        "af1": "visa senaste uppladdningar",
+	},
```
Bevinsky, line 554: I think I'm missing a few strings here, but I don't know where they are coming from.

icxes: ag1 was missing from tl.js but should be there now
https://github.com/9001/copyparty/commit/c2cee222bdc3ae836ca1adb52f2af6381e8a5a1f#diff-759e097addf6e21177148e72a8b1763c84506b38e1614b99cade442eeaab4831

copyparty/web/browser.js
chinatsu: could probably change the phrasing to something like "jump to the directory above", but in swedish?

Bevinsky: It could work if it's an action like here, but if it's in reference to just "parent folder" then it might sound weird... I'll have a look to see if it works in most contexts.

Bevinsky: I think we'll just keep this for now unless someone really objects to it.

copyparty/web/browser.js
```diff
@@ -8185,6 +8185,635 @@ var Ls = {
...
+		"ul_btnu": "L A D D A  U P P",
+		"ul_btns": "S Ö K",
+
+		"ul_hash": "hashar",
+		"ul_send": "skickar",
+		"ul_done": "klar",
+		"ul_idle1": "inga uppladdningar har köats",
+		"ut_etah": "average &lt;em&gt;hashing&lt;/em&gt; speed, and estimated time until finish",
+		"ut_etau": "average &lt;em&gt;upload&lt;/em&gt; speed and estimated time until finish",
+		"ut_etat": "average &lt;em&gt;total&lt;/em&gt; speed and estimated time until finish",
```
Bevinsky: Oops, I forgot these.

Bevinsky:
I would prefer if someone went through this, but there don't seem to be any other Swedes on the internet.

@9001 You can probably merge this unless there's something really egregious I've missed.

9001:
najs! tack så mycket kompis :>

PR #555, opened 2025-08-10
fix PRTY_CONFIG not reading global flags from the config

icxes:

the systemd services are currently broken on the arch package since they use PRTY_CONFIG to get the configuration path, and PRTY_CONFIG currently doesn't seem to read anything in the [global] section of the configuration file at all (though it seems to get the account and volume stuff just fine)

after approximately nine gorillion hours of debugging i came up with this which seems to work, though i'm still not entirely sure why


9001:
thx, ye this works

PR #560, opened 2025-08-10
Add missing translated string in up2k.

Bevinsky:
The translation for the string used in up2k for when an upload is sending data was not actually used.


9001:
nice, thx o/

Issue #234, opened 2025-07-28
Ebook metadata support


Scotsguy:
**Describe the idea / solution you'd like**
Similar to how music files get their tags indexed, I'd love if copyparty would also parse ebooks, specifically .epub (Novels), and maybe .cbz (Comics, Manga).

Epub is essentially XHTML in a .zip, and has an [extensive w3c standard](https://www.w3.org/TR/epub-33/).
.cbz ("Comic Book Zip") is little more than image files in a .zip (or .tar for the little-used .cbt format), but there's a [de-facto standard of a ComicInfo.xml](https://github.com/anansi-project/comicinfo) file containing metadata.

Also great (for the grid view, in particular) would be generating thumbnails. Epub has a cover-image property for this, for .cbz everyone just uses the first page I think (I looked at [koreader](https://github.com/koreader/koreader))

**Describe any alternatives you've considered**
[Kavita](https://github.com/Kareadita/Kavita), [Komga](https://github.com/gotson/komga) and [calibre-web](https://github.com/janeczku/calibre-web) exist, but they're all more heavy-weight than I want and also abstract files away from you, which I also don't want

A web-based ebook reader is also overkill - personally, I have an ereader for that.

Scotsguy:
I've just realized that thumbnails already get generated for .cbz. However, it seems to select pages based on the zipfile order, which is not how ereaders usually pick pages (they do it alphabetically)

Made a PR: #246

w0nd3r365:
Nice request. I hope ed doesn't get overwhelmed with having new feature requests like this, since I think it requires significant understanding of the relevant standard specs, which can be many pages. I'd help out here if I had the skills/knowledge, but I don't currently do.

I dislike Calibre's ebook mgmt system, and way back I sought some local OPDS solution for use with other ebook readers, but never setled on any. If possible, copyparty can add OPDS catalog support. Thanks, I just found out about copyparty.
- [Awesome OPDS](https://github.com/opds-community/awesome-opds)
- [Official OPDS webpage](https://opds.io/)

SpazzTL:
I second this, I primarily use copyparty to host my epubs for my ebook readers, and would LOVE to be able to see covers render correctly, even without metadata.

philips:
@Scotsguy Can this be closed? It seems like the feature you want exist now.

Scotsguy:
> [@Scotsguy](https://github.com/Scotsguy) Can this be closed? It seems like the feature you want exist now.

[...]

PR #561, opened 2025-08-10
EPUB Thumbnailing support

Scotsguy:
image

Supports both EPUB 3 and EPUB 2. The latter doesn't have a formal standard for cover images, but this implementation seems to be quite common.

HAVE_SECURE_EXPAT is required because EPUB is XML, and parsing XML is insecure otherwise. This might mean some old versions of python can't get EPUB thumbnails, except for force-enabling it. I'd like some more eyes on that, if possible.

cc #234 


Scotsguy:
Just tested it, and Kobo EPUBs (.kepub.epub) also just work :tm:

9001:
sweet! thanks :grin:

PR #576, opened 2025-08-12
Simplify reverse proxy path handling in redirect()

### LINKED ISSUE #306: Incorrect redirect when using reverse proxy on a subdomain
### Describe the bug
When browsing folders on the server, it will incorrectly redirect you to the website's root url, instead of redirecting you to the root folder.

### To Reproduce
1. Setup Copyparty with `rp-loc: /copyparty`, in the config.
2. Setup Nginx with the `location /copyparty/` block, and `proxy_pass http://cpp_uds/copyparty/;` inside, using the sample config the template. The result will have my website at `https://websiteurl`, and Copyparty at `https://websiteurl/copyparty/`
3. When using Copyparty, pressing [prev], [up] or [next], will eventually land you on `https://websiteurl`.

### Expected behavior
When browsing folders, I expect not to be kicked to the root url.

### Screenshots
https://github.com/user-attachments/assets/12bb7a66-0a3e-48f4-ae80-3def04639568

### Server details
* **server OS / version:** Debian 12
* **what copyparty did you grab:** sfx
* **how you're running it:** systemd
* run copyparty with `--version` and grab the last 3 lines (they start with `copyparty`, `CPython`, `sqlite`) and paste them below this line:
```
copyparty v1.18.6 "logtail" (2025-07-28)
  CPython v3.11.2 on Linux64 6.1.140 [GCC 12.2.0]
   sqlite 3.40.1*1 | jinja 3.1.2 | pyftpd 1.5.10 | tftp 0.4.0
```
* **copyparty config-file:** 
```
  i: unix:770:www:/dev/shm/party.sock
  xff-src: lan
  rp-loc: /copyparty
```
* **nginx config-file:** 
```
upstream cpp_uds {
        # alternative 2: unix-socket, aka. "unix domain socket";
        # 5-10% faster, and better isolation from other software,
        # but there must be at least one unix-group which both
        # nginx and copyparty is a member of; if that group is
        # "www" then run copyparty with the following args:
        # -i unix:770:www:/dev/shm/party.sock
        server unix:/dev/shm/party.sock fail_timeout=1s;
        keepalive 1;
}

# Default server configuration
#
server {
        listen 80 default_server;
        listen [::]:80 default_server;

        # SSL configuration
        #
        # listen 443 ssl default_server;
        # listen [::]:443 ssl default_server;
        #
        # Note: You should disable gzip for SSL traffic.
        # See: https://bugs.debian.org/773332
        #
        # Read up on ssl_ciphers to ensure a secure configuration.
        # See: https://bugs.debian.org/765782
        #
        # Self signed certs generated by the ssl-cert package
        # Don't use them in a production server!
        #
        # include snippets/snakeoil.conf;

        root /var/www/html;

        # Add index.php to the list if you are using PHP
        index index.html index.htm index.nginx-debian.html;

        server_name _;

        location / {
                # First attempt to serve request as file, then
                # as directory, then fall back to displaying a 404.
                try_files $uri $uri/ =404;
        }

        # pass PHP scripts to FastCGI server
        #
        #location ~ \.php$ {
        #       include snippets/fastcgi-php.conf;
        #
        #       # With php-fpm (or other unix sockets):
        #       fastcgi_pass unix:/run/php/php7.4-fpm.sock;
        #       # With php-cgi (or other tcp sockets):
        #       fastcgi_pass 127.0.0.1:9000;
        #}

        # deny access to .htaccess files, if Apache's document root
        # concurs with nginx's one
        #
        #location ~ /\.ht {
        #       deny all;
        #}

        # --- Copyparty Reverse Proxy ---

        # This block handles ONLY requests for the /copyparty/ path.

        location /copyparty/ {
                # Pass the request to the Copyparty service.
                # The trailing slashes here are critical for correct path mapping.
                proxy_pass http://cpp_uds/copyparty/;
                proxy_redirect off;

                # disable buffering (next 4 lines)
                proxy_http_version 1.1;
                client_max_body_size 0;
                proxy_buffering off;
                proxy_request_buffering off;
                # improve download speed from 600 to 1500 MiB/s
                proxy_buffers 32 8k;
                proxy_buffer_size 16k;
                proxy_busy_buffers_size 24k;

                proxy_set_header   Connection        "Keep-Alive";
                proxy_set_header   Host              $host;
                proxy_set_header   X-Real-IP         $remote_addr;
                proxy_set_header   X-Forwarded-Proto $scheme;
                proxy_set_header   X-Forwarded-For   $proxy_add_x_forwarded_for;
                # NOTE: with cloudflare you want this X-Forwarded-For instead:
                #proxy_set_header   X-Forwarded-For   $http_cf_connecting_ip;
        }
}

# default client_max_body_size (1M) blocks uploads larger than 256 MiB
client_max_body_size 1024M;
client_header_timeout 610m;
client_body_timeout 610m;
send_timeout 610m;
```

### Client details
Tested on multiple clients, same behavior.

### Additional context
`www` group is configured correctly.

gkdis6:
Remove redundant SRS check since SRS paths always start with SR. The first condition (SRS) is redundant because SRS = SR + "/", so any path starting with SRS will also start with SR.

Closes #306(9001/copyparty#306)

### DIFF
--- a/copyparty/httpcli.py
+++ b/copyparty/httpcli.py
@@ -1106,7 +1106,18 @@ class HttpCli(object):
         status: int = 200,
         use302: bool = False,
     ) -> bool:
-        vp = self.args.SRS + vpath
+        if self.args.R:
+            if vpath.startswith(self.args.SR):
+                vp = vpath
+            elif vpath.startswith("/"):
+                vp = self.args.SR + vpath
+            else:
+                vp = self.args.SRS + vpath
+        else:
+            if vpath.startswith("/"):
+                vp = vpath
+            else:
+                vp = "/" + vpath if vpath else "/"
         html = self.j2s(
             "msg",
             h2='<a href="{}">{} {}</a>'.format(


9001:
thanks, but the redirect function is never supposed to be called with a destination which includes the proxy location; if it is, then that is a bug.

with that in mind, this change would be quite the opposite of a simplification :-)

the solution to #306 is to erase the source of these requests in the first place, and I will do this.

PR #579, opened 2025-08-12
warn if failed to import pyvips

mati1210:
pyvips uses `ffi.callback()`, which [is not supported on on some systems](https://cffi.readthedocs.io/en/latest/using.html#callbacks), causing copyparty quietly fail to import it.

## Testing
```sh
$ systemd-run -t -p MemoryDenyWriteExecute=true copyparty
WARNING:root:libvips found, but failed to load: Cannot allocate write+execute memory for ffi.callback(). You might be running on a system that prevents this. For more information, see https://cffi.readthedocs.io/en/latest/using.html#callbacks
```
<sub>This PR complies with the DCO; https://developercertificate.org/</sub>

### DIFF
--- a/copyparty/th_srv.py
+++ b/copyparty/th_srv.py
@@ -113,14 +113,17 @@ except:
 
 try:
     if os.environ.get("PRTY_NO_VIPS"):
-        raise Exception()
+        raise ImportError()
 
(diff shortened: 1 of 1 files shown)


9001:
thx!

PR #584, opened 2025-08-12
feat: added cr3 to raw formats list

djpiper28:


9001:
have you checked if this is actually supported by libraw?

djpiper28:
The camera I use (that makes CR3 files) is in their support camera list. I will sanity this further.

9001:
ah that's probably good then, worst case you'll get corrupted thumbnails if it's not actually supported I suppose :+1: 

thx!

djpiper28:
yeah libraw has CR3 symbols in its source code so I think that all is good

PR #640, opened 2025-08-19
Update browser.js - Found a typo in Dutch language

DeStilleGast:
So I was using the software noticed a part that I could click and noticed a stupid typo I made, also changed the context to be better fitting


9001:
thanks :>

PR #703, opened 2025-08-27
build nix package from repo state instead of a pinned source-tarball

D4ndellion:
Having packages pinned to a release like this is pretty unexpected for a nix user.

`nix build github:9001/copyparty/<ref>` should build the package referred to by that ref (a tag or commit hash or branch or whatever) not the last stable version at that point in time

I guess it was done like this because the package wasn't built from source anyways, but that's since been fixed :smiley: 


9001:
I'm totally green on how the nix stuff works, but will this still be able to obtain the third-party dependencies which are built with [deps-docker/Makefile](https://github.com/9001/copyparty/blob/hovudstraum/scripts/deps-docker/Makefile) and included in the tar.gz of each release? For example `copyparty-1.19.5/copyparty/web/deps/marked.js.gz ` and the other files in web/deps that is --

also pinging @chinponya since he was the mastermind behind the initial package and tends to have good ideas :>

D4ndellion:
Ah, I didn't test the markdown editor, but you're right that it fails. It doesn't have any files in `web/deps` (other than the `__init__` file and whatnot)

That means the nix package stuff isn't actually from source though.. So we have to fix that and package those deps properly

chinponya:
Your assessment is correct @9001. This version will not contain web deps.
This issue was brought up previously in https://github.com/9001/copyparty/pull/253#issuecomment-3145991534.
Ideally we would replicate their build process in nix, but that carries additional maintenance burden, which is not great when the only person who will mess with dependencies doesn't know nix. At that point the nix package might be overstepping its boundaries.
Though maybe now with more eyes on the project this is viable. This will still require a brave soul willing to do the first pass porting [this beautiful build script](https://github.com/9001/copyparty/blob/hovudstraum/scripts/deps-docker/Dockerfile).

PR #717, opened 2025-08-28
Set org.opencontainers.image.source label correctly in all dockerfiles

EmilyxFox:
Opencontainers annotations are a bit vaguely described and up to interpretation (at least i think `org.opencontainers.image.source` is). The spec is described [here](https://github.com/opencontainers/image-spec/blob/main/annotations.md). I do actually think you're supposed to be using actual annotations instead of labels, but I don't know how to do that with podman and your whole build flow, and I've seen plenty of people using labels instead.

The correct source label matters to me because renovate uses it to find changelogs which is quite fancy, but it can't find this repo and just links to alpine instead which is quite silly.

If you're wondering if image.url and image.source should be the same then I also looked into that and its fine. image.url could point to a website or something, but this repo is as close to a website that copyparty has :)

### DIFF
--- a/scripts/docker/Dockerfile.ac
+++ b/scripts/docker/Dockerfile.ac
@@ -1,7 +1,7 @@
 FROM    alpine:latest
 WORKDIR /z
 LABEL   org.opencontainers.image.url="https://github.com/9001/copyparty" \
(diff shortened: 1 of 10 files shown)


9001:
good catch, thanks o/

PR #745, opened 2025-09-01
ftpd: Use implicit FTPS instead of explicit

hw0lff:
This PRs intent is to show it is possible to use implicit FTP over TLS with pyftpdlib.

Implementation found in https://github.com/giampaolo/pyftpdlib/commit/edbc9ccaaac7beccd4a2c0ea5ffb56a9512ba02c

### DIFF
--- a/copyparty/ftpd.py
+++ b/copyparty/ftpd.py
@@ -564,6 +564,16 @@ try:
     class SftpHandler(FtpHandler, TLS_FTPHandler):
         pass
 
+    class FtpsImplicitHandler(FtpHandler, TLS_FTPHandler):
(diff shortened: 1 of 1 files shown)


9001:
thanks but this will require a fair amount of work to integrate, it is not serving any purpose as-is, i plan to return to this when i get the chance to, it is fine to not rebase this until that time

hw0lff:
I just apply this patch to my personal version of copyparty.
By no means I desired to put pressure on you by rebasing.
I'll just do this on a different branch from now, so that this PR (and you) will not get spammed.
I'm sorry!

PR #749, opened 2025-09-02
Add pwa support, allow pwa-ing sub-folders

hw0lff:
Announcing a [webmanifest](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest) makes the website installable as a PWA.


9001:
thanks but this is too expensive as-is, a different approach will be necessary but im not sure what yet, i plan to return to this when i get the chance to, perhaps december or early next yera, it is fine to not rebase this until that time

hw0lff:
oh, sorry i didn't want to put pressure on you by rebasing.
I just patch copyparty with this for myself.
I'll just do this on a different branch from now, so that this PR (and you) will not get spammed with the rebases.

ilotoki0804:
Can this be merged? IMHO the implementation is simple and clean, and it will greatly benefit many users.

tilse:
@hw0lff i have separately worked on PWA support for UI-V1.5 https://github.com/9001/copyparty/pull/1453 , maybe you could have a look at implementing folder-level PWA links in there? my implementation comes with web-share target support so it goes a little further with OS integration
i'd also suggest inverting the argument logic, because imo PWAs should be supported by default

hw0lff:
@tilse I looked at #1453 just for a few minutes.
From what I can see, the manifest.json is currently a static file there.
In my current implementation it is generated dynamically.
In my understanding of copypartys code, it needs to be generated at request time to set the `start_url` parameter correctly.

Back when I wrote this, I couldn't figure out where to properly intercept the web request for the manifest.json so I added the `?pwa` URL parameter as a workaround.

PS/Side note: Only the manifest.json is a hard requirement for a PWA. A service worker is optional and only necessary for additional functionality.

Issue #700, opened 2025-08-26
Nixos repeatable options not supported


danbulant:
In the [nix file](https://github.com/9001/copyparty/blob/hovudstraum/contrib/nixos/modules/copyparty.nix), all array values are transformed into comma separated, which works for the example with `p`, but not for repeatable options, as nix doesn't support repeated keys in objects.

This could be easily set via [`listsAsDuplicateKeys`](https://noogle.dev/f/lib/generators/toKeyValue), except that would in turn break comma separated values.

Either copyparty itself should choose one of those (or support both for some things? For example always support repeating the command and just merge it into comma separated values?) or the nix file should have a whitelist of options that are repeatable. This for example impacts `ipu` which now crashes with error that it must be CIDR=name and not comma separated CIDR1=name1,CIDR2=name2 etc.

Or just allow setting custom command line options for copyparty in the service definition.

Edit for anyone finding this:
`globalExtraConfig` was added that adds custom string to [global]:

```
services.copyparty.globalExtraConfig = ''
  mime: qoi=image/x-qoi
  mime: adf=application/x-amiga-disk-format
  ipu: 172.16.24.0/24=dave
  ipu: 172.16.27.0/24=ben
'';
```

Thanks  a lot!

9001:
some of the comma-separated options are non-repeatable because they have a default-value which should get discarded and replaced with the user-provided value(s); the ones which are repeatable all have the default of being unset. Right now, things get a bit messy when a default value is assigned to a list, because perhaps the user intentionally wanted it to be blank. I'm sure this could be improved with some thinking and refactoring, but right now it is what it is...

PR #751, opened 2025-09-02
Added globalExtraConfig option to the NixOS Module

xvrqt:
Added in an option, 'services.copyparty.globalExtraConfig', with default value and description to the NixOS Module. The option type is 'str' and the default value is the empty string.

This string is appended verbatim to the [global] section of the config. This allows the use of settings which rely on repeated values to be correctly used. For example, the: 'ipu: 255.255.255.1/32=user' key which allows automatic sign in for users of a CIDR subnet. Because attribute sets in Nix must have unique keys, it is not possible to set more than one CIDR subnet/user pair.

'extraConfig' options are common in many Nix modules, providing a simple and familiar configuration which is also flexible for future options.

### DIFF
--- a/contrib/nixos/modules/copyparty.nix
+++ b/contrib/nixos/modules/copyparty.nix
@@ -135,7 +135,7 @@ in
     globalExtraConfig = mkOption {
       type = types.str;
       default = "";
-      description = "Appened to the end of the [global] section verbatim. This is useful for flags which are used in a repeating manner (e.g. ipu: 255.255.255.1=user) which can't be repeated in the settings = {} attribute set.";
+      description = "Appended to the end of the [global] section verbatim. This is useful for flags which are used in a repeating manner (e.g. ipu: 255.255.255.1=user) which can't be repeated in the settings = {} attribute set.";
     };
 
     accounts = mkOption {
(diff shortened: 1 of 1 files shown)


9001:
nice! thanks for finding a workaround for #700.

does this also make it possible to specify something like the following (multiple options with multiple values each)?

```yaml
[global]
  mime: qoi=image/x-qoi
  mime: adf=application/x-amiga-disk-format
  ipu: 172.16.24.0/24=dave
  ipu: 172.16.27.0/24=ben
```

btw, i plan to help mitigate this issue by making more of the repeatable options also accept comma-separated values, but not all of them (since some may have values which themselves contain a comma).

small nitpick: typo "Appened"

xvrqt:
@9001 Yes, it would also allow for specifying your example. 
You could put the following into your Nix config:
```nix
services.copyparty.globalExtraConfig = ''
  mime: qoi=image/x-qoi
  mime: adf=application/x-amiga-disk-format
  ipu: 172.16.24.0/24=dave
  ipu: 172.16.27.0/24=ben
'';
```

9001:
sweet!

@chinponya , any objections on this?

chinponya:
@9001 nah, it seems fine to me. Probably the best way to handle it at the moment.
I feel like the whitelist approach mentioned in the linked issue would set a bad precedent for the module trying to be too smart (and the list would need to be kept up to date) - at that point we'd be better off going all in, have copyparty generate some sort of 'schema' nix could use and generate a proper type checked options list.

PR #779, opened 2025-09-08
OPDS Support

### LINKED ISSUE #498: OPDS feed support for select folders
This suggestion relates to OPDS if you have no idea what it is you can find more here [Wikipedia](https://en.wikipedia.org/wiki/Open_Publication_Distribution_System). In short what it does is make a feed available to know the contents of a server/library. This is used by most libraries to let their users browse and access their books.

I, like many others, would like to host my own private library of all the books I have, and would like to have access to it from anywhere. The current options for doing this are either to keep all your books in one storage device (such as a usb key) and always have it with you. Or use software like Calibre that manages the library and can publish an OPDS feed, which you can then add to your e-reader app/device such as Moon+ Reader on android or Thorium on Linux/Windows.

From what I have seen while looking around for what exactly OPDS is I can't see why any file server would not be able to do it. 
In my mind this feature would allow you to mark a folder as a bookcase/library and you would then get a link to the OPDS feed that you can the put into your reader of choice.

The only alternative is to host Calibre (or an alternative) on your server, I simply don't see why Copyparty could not do it. From my looking around it seems to boil down to:
Read metadata tags -> Create OPDS feed -> Have that feed available on some link -> Profit

Scotsguy:
- [x] Can browse directories thru opds
- [x] Can see thumbnails/covers
- [x] Server Config
- [x] Permissions checks (browse permission)
- [x] don't template XML in a .html file
- [ ] ~~KOReader seems to not get the correct filenames, maybe check content-disposition~~ file downloads generally dont have a content-disposition: attachment; gonna call this out of scope for now
- [ ] ~~(maybe) other metadata. would probably need better metadata support first.~~ this can wait until after we collect that metadata in the first place
- [x] documentation

Tested with KOReader and Moon+ Reader (Android). The latter has an (apparently) rather common bug with basic-auth.

Closes #498 

Shoutout to the [W3C Atom feed validator](https://validator.w3.org/feed/#validate_by_input)


README.md
```diff
@@ -1039,6 +1039,27 @@ url parameters:
...
+for example: `https://copyparty.example/books/?opds`.
+
+to log in with a password, enter it into either of the username or password fields in your client.
+
+- if you've enabled `--usernames`, then you need to enter both username and password .
+
+note: some clients (e.g. Moon+ Reader) will not send the password when downloading cover images, which will
+cause your ip to be banned by copyparty. to work around this, you can grant the [`g` permission](#accounts-and-volumes)
+to unauthenticated requests and enable [filekeys](#filekeys) to prevent guessing filenames. for example:
+`-vbooks:books:r,ed:g:c,fk,opds`
```
Scotsguy, line 1058: This could probably be worked around without user intervention by adding `&pw=hunter2` to the hrefs in the OPDS feed, but I'm hesitant of sending the user's password in plaintext, so I didn't implement it. thoughts?

9001:
hoping to take a proper look sunday :+1: 

> file downloads generally dont have a content-disposition: attachment; gonna call this out of scope for now

if a browser receives a content-disposition header, it tends to assume the file should be downloaded rather than displayed inline, so that's not default-enabled, however url-param `?dl` will add one

README.md
9001: yeah probably best to not do that -- I think your reocmmendation of `g` + `fk` is the best approach

Scotsguy: good, then I'll just leave it as is

Scotsguy:
> if a browser receives a content-disposition header, it tends to assume the file should be downloaded rather than displayed inline, so that's not default-enabled, however url-param `?dl` will add one

ah, i didn't see that
with 2969d9ff9243af7d76f10c8132e753871e0df1e0 opds downloads now also get sent a content-disposition
KOReader uses `filename` instead of `filename*` because of course it does, but that's not our bug

copyparty/httpcli.py
```diff
@@ -6646,14 +6668,55 @@ def tx_browser(self) -> bool:
             else:
                 href = quotep(href)
 
+            mime = None
```
9001: rather than having the opds logic inside the `for fn in ls_names:` loop, could we move it down to below the loop, and make the whole thing a single conditional? this would move it out of the hot-path and help for performance;

```python
for fn in ls_names:
    # unmodified original code
for d in dirs:
	d["name"] += "/"
dirs.sort(key=itemgetter("name"))

# the remaining opds stuff can go here, modifying the `dirs` / `files` lists:
if is_opds:
    # exclude files which don't match --opds-exts (originally done on line 6647);
	if self.args.opds_exts:
		files = {x for x in files if x["name"].rsplit(".", 1)[-1] in self.args.opds_exts}
	# insert attributes into each file
    for item in files:
		href = item["href"]
		href += ("&" if "?" in href else "?") + "opds"
		# [...]
	# then into each file
	for item in dirs:
		href = item["href"]
		href += ("&" if "?" in href else "?") + "opds"
		# [...]
```

i'm fine with duplicaitng the code for generating the iso8601 in each of those forloops (files/dirs), think that is best for performance anyways

copyparty/__main__.py
```diff
@@ -1432,6 +1432,10 @@ def add_smb(ap):
     ap2.add_argument("--smbvv", action="store_true", help="verboser")
     ap2.add_argument("--smbvvv", action="store_true", help="verbosest")
 
+def add_opds(ap):
+    ap2 = ap.add_argument_group("OPDS options")
+    ap2.add_argument("--opds", action="store_true", help="enable opds -- allows e-book readers to browse and download files (volflag=opds)")
+    ap2.add_argument("--opds-allowed", metavar="T,T", type=u, default="epub,cbz,pdf", help="file formats to list in OPDS feeds; leave empty to show everything (volflag=opds_allowed)")
```
9001: could we rename this to --opds-exts? This would avoid the first column in the helptext from becoming wider, which helps for readability imo

Scotsguy:
Hi, did you get a chance to review the PR yet?

9001:
realizing I don't know how to GitHub -- guessing you're not seeing these? 
Screenshot_20250922-150759
Screenshot_20250922-150833~2

Scotsguy:
You've started a review, but since you're in review mode those aren't published yet
there should be a button called publish review or similar to make them public

9001 (changes requested):

9001:
👴👍

copyparty/httpcli.py
Scotsguy: Does self.args.opds_exts automatically merge/override when volflags are present?

Scotsguy: Seems like it doesn't, I put back in the fallback logic

copyparty/__main__.py
Scotsguy: Sure, done

Scotsguy:
Small bug that has popped up:
when starting up, copyparty logs all volflags for all volumes
if `opds_exts` is set, it'll log them with commas between every character, like so:
```
18:17:10.246 up2k                  /opds/ opds opds_exts:e,p,u,b,,,c,b,z shr_who:no

```

happens here:
https://github.com/9001/copyparty/blob/e09f3c9e2c3dccf8f3912539e04dd840b10b51ee/copyparty/up2k.py#L1164

9001:
good stuff! just a few nitpicks I tweaked before merging, and you accidentally pointed out an optimization opportunity too :>

philips:
@Scotsguy Thanks for doing this! I would like to add OpenSearch support so I can search my volume from OPDS. Did you investigate that at all? Just curious if you did before I start hacking.

My plan was just basic filename search: https://specs.opds.io/opds-1.1.html#Search

Scotsguy:
i looked into opensearch, yes, but then didn't implement it because it seemed that adapting opensearch to copyparty's native search API would be a pain. feel free to implement it, I'll probably have a look when you're done :)

fyi: you linked to opds 1.1, but i implemented this PR while looking at the specs for 1.2. I'm not sure what the differences are, but you should probably look at 1.2 regardless

PR #787, opened 2025-09-10
Esperanto translation

slashdevslashurandom:
This is, admittedly, a rather quick, but complete translation of the copyparty interface (or at least the parts that are listed in `browser.js`) into Esperanto. So far, I didn't notice any serious breakages when testing, but I haven't tested all (or most) of copyparty's functionality yet.

For purposes of grammar: Esperanto uses the same number pluralization rules as English, one string for 1 of something (1 dosiero), one for any other number (0 dosieroj, 10 dosieroj, 21 dosieroj, etc.)

One of the translated strings also introduces an emoji where there isn't one, the "uncache" button doesn't have a short enough translation that won't be confusing, so I added a trashcan icon next to an abbreviation for "cache" instead.

I'll continue updating this branch if I see that anything needs fixing. There are still some stylistic decisions I haven't completely settled on, either.


9001:
thanks for adding this! I didn't have time to merge it for today's release but I will before the next, probably in 1 to 3 weeks. 

PS: don't worry about keeping it up to date with any changes to the javascript file, I'll fix that as necessary

PPS: there are also some strings in splash.js if you want to translate those, you can grab a copy of the english ones from https://github.com/9001/copyparty/blob/hovudstraum/scripts/tl.js

9001:
soryr i didnt catch this initially  but could you move the translation up between deu and fin so it is alhpabetic?

slashdevslashurandom:
Done.

9001:
thanks :grin: :+1:

PR #789, opened 2025-09-10
case-insensitive search for non-ascii characters

### LINKED ISSUE #715: Case-insensitive search does not work for non-ASCII characters
### Description
Search is usually case-insensitive, but it does not work for characters outside of ASCII.

### To Reproduce
1. Have two files named `å1.txt` and `Å2.txt`.
2. Search for `å` in filenames.

### Expected behavior
Get two files in search results since they both have `å` just with different cases.

### Server details
* **server OS / version:** Windows 11 Pro 23H2
* **what copyparty did I grab:** from uv
* **how I'm running it:** I type `uvx copyparty` in command prompt
* `copyparty --version`:
  * copyparty v1.19.6 "usernames" (2025-08-27)
  * CPython v3.13.6 on Windows64 10.0.22631 [MSC v.1944 64 bit (AMD64)]
  * sqlite 3.49.1*1 | jinja 3.1.6 | pyftpd (None) | tftp (None)
* **copyparty arguments and config-file:** `-p 80 -e2dsa` to enable search

### Additional context
I looked in the code how searching is done. The culprit is sqlite. Its developers [intentionally](https://www.sqlite.org/faq.html#q18) don't do case-insensitive matching for Unicode characters. I also can confirm that with database viewer. For example, here is the index of files:

Image

When I run the search query for a Unicode character, I get only the rows with the same case:

Image

Since it is how sqlite is supposed to work, I'm afraid fixing this issue will be too much hassle for little comfort. Python's lower() function works correctly for all Unicode that I'm aware of. Therefore, it can be used to fix the issue, but I don't know how to implement it in copyparty (all I managed to do is to break the database). The same sqlite answer mentiones the possibility of overloading their default functions for working with letters case, but I have no idea how that works. Adding another column with lowered filenames should also work, but most users will not want to have increased database size just for this feature, so it can be made optional with a config toggle.

This issue should have little impact for many languages, but for Cyrillic languages it makes the search case-sensitive.

km-clay:
fixes case-insensitive searches for non-ascii characters by making use of python's `casefold()` method for strings.

closes #715 

pr1

### DIFF
--- a/copyparty/u2idx.py
+++ b/copyparty/u2idx.py
@@ -53,12 +53,17 @@ class U2idx(object):
             self.log("your python does not have sqlite3; searching will be disabled")
             return
 
+        if self.args.srch_icase:
+            self._open_db = self._open_db_icase
+        else:
+            self._open_db = self._open_db_std
+
         assert sqlite3  # type: ignore  # !rm
 
         self.active_id = ""
         self.active_cur: Optional["sqlite3.Cursor"] = None
         self.cur: dict[str, "sqlite3.Cursor"] = {}
-        self.mem_cur = self._connect_db(":memory:", check_same_thread=False).cursor()
+        self.mem_cur = sqlite3.connect(":memory:", check_same_thread=False).cursor()
         self.mem_cur.execute(r"create table a (b text)")
 
         self.sh_cur: Optional["sqlite3.Cursor"] = None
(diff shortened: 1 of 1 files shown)


9001:
nice! really slick solution :>

I hope to give this a spin + merge on sunday :+1:

9001:
made some adjustments to also include filepaths into this, but excluded tags for now -- I think those deserve to be indexed twice, with both original and lowercased values, which should help preserve performance. The current plan is to insert a second copy of each tag with the keyname suffixed with `.lc` or perhaps `:lc` (just an idea, I believe one of those will cause issues), I'll get back to that at some point.

Thanks again o/

PR #835, opened 2025-09-21
fix: rootless podman/docker not building correctly on NixOS

bartoostveen:
Title.

Talked about this on Discord. Without this, tar will refuse to unpack because the target user does not exist on some platforms (NixOS for example).

Adding `--no-same-owner` at least should fix this on all platforms. According to `tar --help`:

```
      --no-same-owner        extract files as yourself (default for ordinary
                             users)
```


9001:
good stuff, thanks o/

PR #840, opened 2025-09-22
versus.md: add links with tooltips to the comparison matrices

guano:
Just added github markdown link references to the other softwares' web sites. This way the links could be added into all of the matrices and the hover tooltip will show the name of the software.

I debated having the link go to each software's review section in the same document instead of the web site, but links within the same document don't allow the back button to go back to where you were, so I decided it wasn't as useful.


9001:
sweet! cool trick, thanks :>

PR #860, opened 2025-09-27
thumb/epub: handle missing covers

Scotsguy:
Some epub files don't have covers at all. In this case, write an empty file as the thumbnail.
Also, some refactoring to allow au_unpk to return None without going through an Exception.
Not entirely sure I understood the worker logic correctly, I might've broken something during the refactor. Would appreciate a thorough review on that part :)


9001:
went with a slightly different approach, but thx a lot for the ideas :>

PR #865, opened 2025-09-28
Working basic example of authentik auth integration with copy party

KuroIvy:
Updated authentik docker example with working example stacks and updated README to better explain how to get setup.  I did my best to make it a basic level of secure and simple. If it can be improved please mention how! 😁  The key was specifying the xff-src, and using the correct authentik default headers.

### DIFF
--- a/docs/examples/docker/idp-authentik-traefik/README.md
+++ b/docs/examples/docker/idp-authentik-traefik/README.md
@@ -1,8 +1,7 @@
 > [!WARNING]  
 > I am unable to guarantee the quality, safety, and security of anything in this folder; it is a combination of examples I found online. Please submit corrections or improvements 🙏
 
-> [!WARNING]  
-> does not work yet... if you are able to fix this, please do!
+This example should be enough to get things working. I have confirmed this basic config personally. User creation and management work, however once a user is passed through to copy party, manual editing of the config file will be needed to control new users. i.e. anyone you let through your SSO portal will be granted basic/default settings and permissions, unless manually configured otherwise. 
 
 this is based on:
 * https://goauthentik.io/docker-compose.yml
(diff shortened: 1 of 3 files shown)


xfallme:
Just adding to this, as someone how has struggled quite a lot with Authentik in a homelab environment:

- It is always a good idea to run the latest release version of Authentik, at the moment of writing its 2025.10.0.
The current up-to-date docker compose by Authentik can always be found here: https://docs.goauthentik.io/docker-compose.yml 
- Pinning traefik to the latest label _might_ be a recipe for disaster and definitely shouldn't be the value set in an example, nobody wants their reverse proxy to break because of an update (in my experience that's not a fun time)

@KuroIvy @9001 Is this still active?

9001:
this is currently blocked by #943 ; I'm not comfortable making/merging authentication-related changes without careful scrutiny

xfallme:
I am sorry, didn't want disturb your vacation with some random comment. The "still active" question was aimed at the PR in general. But that conversation can wait!

KuroIvy:
> Just adding to this, as someone how has struggled quite a lot with Authentik in a homelab environment:
> 
> * It is always a good idea to run the latest release version of Authentik, at the moment of writing its 2025.10.0.
>   The current up-to-date docker compose by Authentik can always be found here: https://docs.goauthentik.io/docker-compose.yml
> * Pinning traefik to the latest label _might_ be a recipe for disaster and definitely shouldn't be the value set in an example, nobody wants their reverse proxy to break because of an update (in my experience that's not a fun time)
> 
> @KuroIvy @9001 Is this still active?

Hey thanks for the feedback. I mostly just wanted to provide a fully working example as the one in the current contrib isn’t complete. I can update the PR with your notes no problem. I love SSO and copy party is really cool so I did what I could to make it work so thought I would share. :)

nekitbr:
> * [...]
> * Pinning traefik to the latest label _might_ (is) be a recipe for disaster and definitely shouldn't be the value set in an example, nobody wants their reverse proxy to break because of an update (in my experience that's not a fun time)
> * [...]

Btw: enjoy your vacations!

PR #868, opened 2025-09-28
Allow building current code with flake

shelvacu:
Previously, the flake could only build the pinned (release) version of copyparty. Now it can also build the code in the repo around it, with the exception of webdeps which it copies from the pinned version.

With these modifications the `--version` output of `copyparty-unstable` looks like:

```text
copyparty v1.19.14 "unstable" (2025-09-28)
  CPython v3.12.10 on Linux64  [GCC 14.2.1 20250322]
   sqlite 3.48.0*1 | jinja 3.1.6 | pyftpd 2.0.1 | tftp (None)
```

---

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -12,7 +12,8 @@
     }:
     {
       nixosModules.default = ./contrib/nixos/modules/copyparty.nix;
-      overlays.default = import ./contrib/package/nix/overlay.nix;
+      overlays.default = final: prev:
+        (import ./contrib/package/nix/overlay.nix final prev) // { copypartyFlake = self; };
     }
     // flake-utils.lib.eachDefaultSystem (
       system:
@@ -22,26 +23,26 @@
           config = {
             allowAliases = false;
           };
-          overlays = [ self.overlays.default ];
+          overlays = [
+            self.overlays.default
+          ];
         };
       in
       {
         # check that copyparty builds with all optionals turned on
-        checks.copyparty-full = self.packages.${system}.copyparty.override {
-          withHashedPasswords = true;
-          withCertgen = true;
-          withThumbnails = true;
-          withFastThumbnails = true;
-          withMediaProcessing = true;
-          withBasicAudioMetadata = true;
-          withZeroMQ = true;
-          withFTPS = true;
-          withSMB = true;
+        checks = {
+          inherit (pkgs)
+            copyparty-full
+            copyparty-unstable-full
+            ;
         };
 
         packages = {
           inherit (pkgs)
             copyparty
+            copyparty-full
+            copyparty-unstable
+            copyparty-unstable-full
             ;
           default = self.packages.${system}.copyparty;
         };
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ scripts/docker/*.err
 
 # nix build output link
 result
+result-*
 
 # IDEA config
 .idea/
--- a/contrib/package/nix/copyparty/default.nix
+++ b/contrib/package/nix/copyparty/default.nix
@@ -67,18 +67,64 @@
   # additional dependencies
   extraPythonPackages ? (_p: [ ]),
 
+  # to build stable + unstable with the same file
+  stable ? true,
+
+  # for commit date, only used when stable = false
+  copypartyFlake ? null,
+
+  nix-gitignore,
 }:
 
 let
   pinData = lib.importJSON ./pin.json;
   runtimeDeps = ([ util-linux ] ++ extraPackages ++ lib.optional withMediaProcessing ffmpeg);
+  inherit (copypartyFlake) lastModifiedDate;
+  # ex: "1970" "01" "01"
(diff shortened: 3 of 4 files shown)


9001:
i'm entirely blank on the nixos stuff, so --

@chinponya @toast003 could I have your opinion on this?

also if either of you ever get tired of the pings let me know and I'll stop :p

flake.nix
```diff
@@ -38,10 +33,45 @@
...
         };
+        pkgs = import nixpkgs {
+          inherit system;
+          config = {
+            allowAliases = false;
+          };
+          overlays = [ self.overlays.default ] ++ lib.singleton (final: prev: {
+            # This is here in flake.nix rather than common/packages/nix because it needs to access the flake metadata to get the date of the most recent commit (lastModifiedDate)
+            copyparty-unstable = final.copyparty.overrideAttrs (finalAttrs: previousAttrs: {
+              version = "${previousAttrs.version}-unstable-${self.lastModifiedDate}";
```
chinponya: I believe it is preferred to append the commit hash for untagged versions. In this case `self.shortRev` would work.

flake.nix
```diff
@@ -38,10 +33,45 @@
           withFTPS = true;
           withSMB = true;
         };
```
chinponya: Not caused by this PR but I am now realizing that this list is missing a couple attributes such as `withMagic` and `withTFTP`. It might be worth including here.

flake.nix
```diff
@@ -17,17 +17,12 @@
...
-        };
-      in
-      {
-        # check that copyparty builds with all optionals turned on
-        checks.copyparty-full = self.packages.${system}.copyparty.override {
+        lib = import "${nixpkgs}/lib";
+        inherit (self) lastModifiedDate;
+        year = lib.toIntBase10 (builtins.substring 0 4 lastModifiedDate);
+        month = lib.toIntBase10 (builtins.substring 4 2 lastModifiedDate);
+        day = lib.toIntBase10 (builtins.substring 6 2 lastModifiedDate);
```
chinponya: These helpers exist only for the unstable overlay and are defined at the top-level here. Maybe co-locating them with package overrides would be better?

flake.nix
```diff
@@ -38,10 +33,45 @@
           withFTPS = true;
           withSMB = true;
         };
+        pkgs = import nixpkgs {
+          inherit system;
+          config = {
+            allowAliases = false;
+          };
+          overlays = [ self.overlays.default ] ++ lib.singleton (final: prev: {
+            # This is here in flake.nix rather than common/packages/nix because it needs to access the flake metadata to get the date of the most recent commit (lastModifiedDate)
```
chinponya: We might need to keep all overlays within the `overlay.nix` file, in order to make them available for users who do not use flake-enabled setups (and I know they exist here).

Unfortunately I do not have a good recommendation on how to preserve the date/revision once this segment is extracted out of `flake.nix`.

Maybe we don't actually need it and merely appending `unstable` to the previous release version is good enough?

chinponya (changes requested):
Thanks for working on this. This actually helps more than it appears at a glance, since it allows us to more easily test builds.

The idea has some issues: any updates that involve changes to web deps could leave us with a broken build. I'd imagine these will be fairly rare and it is called "unstable" for a reason. As a stop-gap solution until we start building web deps with nix, I think it will do just fine.

I'm probably bikeshedding here too much.

flake.nix
toast003: I agree, I think just adding `unstable` (or `git`) would be good enough

flake.nix
toast003: Imo that should fixed be in a different pr

flake.nix
shelvacu: This was following https://github.com/NixOS/nixpkgs/tree/master/pkgs#versioning

shelvacu:
Moved everything into `contrib/nix/copyparty/default.nix`, except `flake.nix` passes in `self` via an overlay setting `copypartyFlake`. This way you can build both with and without flakes.

To verify everything builds:

```bash
set -xev
for packageName in copyparty{,-unstable}{,-full}; do
  echo "building $packageName with flakes"
  nix build .#$packageName
  echo "building $packageName with old-style"
  nix-build --expr '(import <nixpkgs> { overlays = [(import ./contrib/package/nix/overlay.nix)]; })' -A $packageName
  echo "fin $packageName"
done
```

toast003 (approved):
Looks good to me.
Personally I would call the unstable build git (`copyparty-git` instead of `copyparty-unstable`), but it's not a big deal

shelvacu:
@9001 is this good to merge?

9001:
I'm still in Japan for another week so haven't had the chance to check yet

9001:
rather, since both of the other nix guys gave it the thumbs-up i'll just go ahead and merge it -- looks fine on first glance and it's not like i understand most of this either way o/

PR #870, opened 2025-09-28
fix(opengraph): video embeds

tsuza:
This seems to fix the inconsistent embedding of videos.


9001:
neat! thanks :>

PR #878, opened 2025-10-01
Add Blu-ray discs to fuzzy file size type

scj643:
Added Blu-ray disc sizes to the fuzzy file size type.


9001:
thx!

Issue #867, opened 2025-09-28
copypaste nixOS module - Setting value for services.copypaste.user throws error on service activation.


2Kaleb:
### Describe the bug
`copypaste.nix` nixOS module - Setting value for `services.copypaste.user` throws error on service activation / new generation fails.

### To Reproduce
List of steps to reproduce the issue, or, if it's hard to reproduce, then at least a detailed explanation of what you did to run into it
In flake configuration
```
services.copypaste.user="kdebre"; # The user "kdebre" exists on the system.
 sudo nixos-rebuild switch --flake
```
Even with
`users.users.kdebre.extraGroups = [ "copyparty" ];`
The error still occurs.

### Expected behavior
Successful switch to new generation and copypaste runs under user "kdebre" in htop.
Being able to only change the user of nixos services without also modifying the group has been possible for me with mainline nixos modules, e.g. `services.jellyfin.user="kdebre";`

### Screenshots
See ExecStartPre `copyparty-pre-start`. Idk anything about that binary.
```
warning: the following units failed: copyparty.service
× copyparty.service - http file sharing hub
     Loaded: loaded (/etc/systemd/system/copyparty.service; enabled; preset: ignored)
     Active: failed (Result: exit-code) since Sun 2025-09-28 16:04:14 CEST; 2s ago
   Duration: 43.491s
 Invocation: ae3bd090e8384cbfad7efe8e59060187
    Process: 249857 ExecStartPre=/nix/store/ciabrl84j4s78qymnzyga6rqydsrf2ys-unit-script-copyparty-pre-start/bin/copyparty-pre-start (code=exited, status=216/GROUP)
         IP: 0B in, 0B out
         IO: 0B read, 0B written
   Mem peak: 1.6M
        CPU: 8ms

Sep 28 16:04:14 servitro systemd[1]: Starting http file sharing hub...
Sep 28 16:04:14 servitro (re-start)[249857]: copyparty.service: Failed to determine group credentials: No such process
Sep 28 16:04:14 servitro (re-start)[249857]: copyparty.service: Failed at step GROUP spawning /nix/store/ciabrl84j4s78qymnzyga6rqydsrf2ys-unit-script-copyparty-pre-start/bin/copyparty-pre-start: No such process
Sep 28 16:04:14 servitro systemd[1]: copyparty.service: Control process exited, code=exited, status=216/GROUP
Sep 28 16:04:14 servitro systemd[1]: copyparty.service: Failed with result 'exit-code'.
[...]

PR #886, opened 2025-10-03
nixos module - allow for setting only the user or group under which the service runs + some readme.md

2Kaleb:
Referring to #867 
Allows to run the service as "user", while keeping the default group as "copyparty".
This configuration is more consistent with available nixos options, so perhaps this can be relevant for eventual mainlining. 
FYI, this is my first ever PR. I used my branch as flake input and the new generation was successful, so I guess it should be fine. 
…

### DIFF
--- a/README.md
+++ b/README.md
@@ -2466,6 +2466,10 @@ copyparty on NixOS is configured via `services.copyparty` options, for example:
 ```nix
 services.copyparty = {
   enable = true;
+  # the user to run the service as
+  user = "copyparty"; 
+  # the group to run the service as
+  group = "copyparty"; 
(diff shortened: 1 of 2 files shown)


chinponya (approved):
Neat. Thanks for fixing this. I don't think anything will break as a result of this. LGTM!

Unrelated but it's weird to see `group` and `groups` in the config next to each other, referring to completely different things, but it can't be helped - they each map directly to either nix conventions or cpp config. Maybe in the future we'd be better off nesting all copyparty config groups within `settings`, though I'm not sure whether introducing a breaking change for that is worth it.

9001:
thx o/

Issue #224, opened 2025-07-28
On-the-fly Video Transcoding for Broader Browser Playback


FireflyTheReal:
Problem:
As someone who is not particularly experienced with video codecs or service side media processing, I'm encountering a common issue: When I serve video files, especially MKV files, through copyparty, they often don't play directly in my web browser. I am aware that this happens because browsers don't natively support containers like MKV (and corresponding codecs).

Solution:
I'd like to suggest adding an optional feature for on-the-fly video transcoding. Since copyparty already uses ffmpeg for things like audio processing and thumbnails, it seems like ffmpeg could be used to convert unsupported video containers into browser friendly formats in real-time. 

Pros: improved user experience and copyparty as an even better media server.
Cons: performance and battery, especially on mobile devices running Termux and maybe security concerns with untrusted video files.
Therefore, this feature should be optional and highly configurable, allowing users to enable it only when needed and adjust quality settings.

Thank you for considering this suggestion!

9001:
this is something I wanted to do a while ago (wow, it's been 4 years...) but the first attempt failed; I added the notes here: https://github.com/9001/copyparty/blob/hovudstraum/docs/hls.html

I do want to give this another shot, **EDIT: but i'm not planning to work on this any time soon!** -- or, if someone else does and is able to pull it off (somewhat along the same lines as described below) then that'd be hella dope --

## what I wanted to do:

when someone requests a file, FFmpeg is used to encode the video into 10-second chunks, and it would assemble the chunks (the existing ones, plus the anticipated ones) into an hls file, which [hls.js](https://github.com/video-dev/hls.js) could play as if it was one contiguous stream

if someone seeks in the video, and the necessary chunks didn't exist in the on-disk cache, another FFmpeg would be created, snapping to the nearest preceeding 10sec boundary, encoding from there, skipping any pre-encoded chunks that are found in the on-disk cache

this avoids wasting cpu, since only one encoding pass will be necessary for each target quality level as selected by the user, of which a list of presets are available (some common resolutions, codecs)

video and audio is encoded into separate files so they can be mixed-and-matched as client needs them to be (most want opus, some can only do aac (hello there apple how are you doing today))

## the problems:

i was having trouble producing video chunks which would play seamlessly; there was noticeable glitching / stuttering when chunks were conjoined for playback, guessing I didn't manage to get FFmpeg to be frame-perfect, which is weird because I did everything according to dts/pts and not frames, so don't think it was anything vfr-related either... ah well

[...]

PR #895, opened 2025-10-05
Add mkv playback support

gyurix:
A simple solution to make Firefox 145+ (currently available as nightly version) and all Chromium based browsers support MKV file playback as long as the audio codec is either AAC, AC3 or FLAC. Unfortunately DTS, E-AC3, DD, DD+ audio codecs as well as switching between multiple audio tracks are still unsupported by the browsers.

### DIFF
--- a/copyparty/util.py
+++ b/copyparty/util.py
@@ -424,7 +424,7 @@ def _add_mimes() -> None:
 application json wasm xml pdf rtf zip jar fits wasm
 image webp jpeg png gif bmp jxl jp2 jxs jxr tiff bpg heic heif avif
 audio aac ogg wav flac ape amr
-video webm mp4 mpeg
+video webm mp4 mkv mpeg
 font woff woff2 otf ttf
 """.splitlines():
         k, vs = ln.split(" ", 1)
@@ -444,7 +444,7 @@ image arw=x-sony-arw cr2=x-canon-cr2 crw=x-canon-crw dcr=x-kodak-dcr dng=x-adobe
 image k25=x-kodak-k25 kdc=x-kodak-kdc mrw=x-minolta-mrw nef=x-nikon-nef orf=x-olympus-orf
 image pef=x-pentax-pef raf=x-fuji-raf raw=x-panasonic-raw sr2=x-sony-sr2 srf=x-sony-srf x3f=x-sigma-x3f
 audio caf=x-caf mp3=mpeg m4a=mp4 mid=midi mpc=musepack aif=aiff au=basic qcp=qcelp
-video mkv=x-matroska mov=quicktime avi=x-msvideo m4v=x-m4v ts=mp2t
+video mkv=mp4 mov=quicktime avi=x-msvideo m4v=x-m4v ts=mp2t
 video asf=x-ms-asf flv=x-flv 3gp=3gpp 3g2=3gpp2 rmvb=vnd.rn-realmedia-vbr
 font ttc=collection
 """.splitlines():


9001:
I'm confused; chrome already supports playing mkv-files without this change -- and I verified on nightly firefox as well; it works there too: https://a.ocv.me/pub/demo/#gf-56b55a16

gyurix:
Is there a different version running on the demo page than the version installed using pip install?

9001:
technically yes, however no relevant changes have been made since mkv support was added [for chrome in 2022](https://github.com/9001/copyparty/commit/f8a31cc24f9ee0e03fd7a39a2e91898b5e241235), and I just checked with the pip version locally, and it seems to work here as well -- what is the rest of your setup; perhaps you have a reverseproxy which is altering the mimetype that copyparty is returning to the browser?

ahmed-abdelrazek:
what about h265 and opus?

gyurix:
This is kinda weird then, because without any config change starting copyparty's pip installed version without my changes triggers download for mkv files in both Chrome and Firefox, while with this change it triggers actual playback for me.
@ahmed-abdelrazek I never tested that.

9001:
> what about h265 and opus?

that's not in our hands to do anything about; we are currently at the mercy of what webbrowsers decide to support 

supporting playback of anything that is not natively supported by browsers is a huge undertaking and will not happen soon ; see #224

9001:
@gyurix are you linking into the lightbox / videoplayer like I did, or hotlinking directly to the file? what the browser decides to do when a file is accessed directly by its url is not something we should tamper with, but it *should* be possible to link to the medaplayer to force inline playback that way

another alternative is to add a `?v` suffix to the url like so: https://a.ocv.me/pub/demo/showcase-subs.mkv?v

gyurix:
I did not setup any linking, adding ?v suffix to the url does the trick, but for .mp4 I did not need to do that, that's why I made this change so I would not need to add ?v to mkv files to play them.

9001:
aight, well the problem is that we'd basically be lying to the browser about what filetype we are sending it -- this may avoid one issue, but it will definitely create many new issues, since other clients expect to be given the true / correct filetype description.

if you really want to do this, then you can override the default mimetypes with the `--mime` option, for example `mime: mkv=video/mp4` in the `[global]` section should mimic your changes. But I strongly recommend not doing this!

gyurix:
> aight, well the problem is that we'd basically be lying to the browser about what filetype we are sending it -- this may avoid one issue, but it will definitely create many new issues, since other clients expect to be given the true / correct filetype description.
> 
> if you really want to do this, then you can override the default mimetypes with the `--mime` option, for example `mime: mkv=video/mp4` in the `[global]` section should mimic your changes. But I strongly recommend not doing this!

I see, can you maybe add info about this to the documentation somewhere, that the --mime option is a potential solution for this?

9001:
I really don't want to recommend this approach at all, because it will create new confusing bugs, and waste a lot of time debugging. However I see that the `?v` suffix is not well-documented so I will mention that.

yarikoptic:
<details>
<summary>FWIW, I do not think this worked for me, adding details just in case (click to expand)</summary> 

added
```patch
ha@hopa:~/proj/home-server/copyparty$ git diff .
diff --git a/copyparty/copyparty.conf b/copyparty/copyparty.conf
index 62052ed..7b00e06 100644
--- a/copyparty/copyparty.conf
+++ b/copyparty/copyparty.conf
@@ -3,6 +3,7 @@
   e2ts   # and enable multimedia indexing
   ansi   # and colors in log messages
 
+  mime: mkv=video/mp4
   # uncomment the line starting with q, lo: to log to a file instead of stdout/journalctl;
   # $LOGS_DIRECTORY is usually /var/log/copyparty (comes from systemd)
   # and copyparty replaces %Y-%m%d with Year-MonthDay, so the

```

but still got 

```
Failed to open 19960320-ChuckBD.mkv

Either the file is corrupt, or your browser does not understand the file format or codec

err#4, DEMUXER_ERROR_NO_SUPPORTED_STREAMS: FFmpegDemuxer: no supported streams
```

for http://192.168.1.7:3923/photoalbum/Me/199x-200x/#gf-b9a5e345,sorthref on

```shell
ha@hopa:~/proj/home-server/photoalbum/Me/199x-200x$ mediainfo 19960320-ChuckBD.mkv
General
Unique ID                                : 263869295965519188710403029166121982635 (0xC68360BACE66510EDBEBA31F94AA9AAB)
Complete name                            : 19960320-ChuckBD.mkv
Format                                   : Matroska
Format version                           : Version 2
File size                                : 746 MiB
Duration                                 : 23 min 0 s
Overall bit rate mode                    : Variable
Overall bit rate                         : 4 532 kb/s
Frame rate                               : 29.970 FPS
Encoded date                             : 2025-10-18 23:42:07 UTC
Writing application                      : MakeMKV v1.18.2 linux(x64-release)
Writing library                          : libmakemkv v1.18.2 (1.3.10/1.5.2) x86_64-linux-gnu

Video
ID                                       : 1
ID in the original source medium         : 224 (0xE0)
Format                                   : MPEG Video
Format version                           : Version 2
Format profile                           : Main@Main
Format settings                          : CustomMatrix / BVOP
Format settings, BVOP                    : Yes
Format settings, Matrix                  : Custom
Format settings, GOP                     : M=3, N=15
Format settings, picture structure       : Frame
Codec ID                                 : V_MPEG2
Codec ID/Info                            : MPEG 1 or 2 Video
Duration                                 : 23 min 0 s
Bit rate mode                            : Variable
Bit rate                                 : 4 142 kb/s
Maximum bit rate                         : 9 334 kb/s
Width                                    : 720 pixels
Height                                   : 480 pixels
Display aspect ratio                     : 4:3
Frame rate mode                          : Constant
Frame rate                               : 29.970 (30000/1001) FPS
Standard                                 : NTSC
Color space                              : YUV
Chroma subsampling                       : 4:2:0
Bit depth                                : 8 bits
Scan type                                : Interlaced
Scan order                               : Top Field First
Compression mode                         : Lossy
Bits/(Pixel*Frame)                       : 0.400
Time code of first frame                 : 00:00:00:00
Time code source                         : Group of pictures header
GOP, Open/Closed                         : Open
GOP, Open/Closed of first frame          : Closed
Stream size                              : 682 MiB (91%)
Language                                 : English
Default                                  : No
Forced                                   : No
Color primaries                          : BT.601 NTSC
Transfer characteristics                 : BT.601
Matrix coefficients                      : BT.601
Original source medium                   : DVD-Video

Audio
ID                                       : 2
ID in the original source medium         : 189 (0xBD)128 (0x80)
Format                                   : AC-3
Format/Info                              : Audio Coding 3
Commercial name                          : Dolby Digital
Codec ID                                 : A_AC3
Duration                                 : 23 min 0 s
Bit rate mode                            : Constant
Bit rate                                 : 384 kb/s
Channel(s)                               : 2 channels
Channel layout                           : L R
Sampling rate                            : 48.0 kHz
Frame rate                               : 31.250 FPS (1536 SPF)
Compression mode                         : Lossy
Stream size                              : 63.2 MiB (8%)
Title                                    : Stereo
Language                                 : English
Service kind                             : Complete Main
Default                                  : Yes
Forced                                   : No
Original source medium                   : DVD-Video
Dialog Normalization                     : -31 dB
compr                                    : -11.02 dB
dialnorm_Average                         : -31 dB
dialnorm_Minimum                         : -31 dB
dialnorm_Maximum                         : -31 dB

Menu
00:00:00.066                             : en:Chapter 01
00:05:13.880                             : en:Chapter 02
00:10:42.708                             : en:Chapter 03
00:16:04.530                             : en:Chapter 04
00:21:22.848                             : en:Chapter 05
```

</details>

9001:
that mkv has codecs which your browser does not support , it is to be expected that many, if not most, mkv files will never play in most web browsers due to this reason , https://github.com/9001/copyparty/pull/895#issuecomment-3369270733

PR #897, opened 2025-10-06
Nix: package web dependencies

toast003:
See https://github.com/9001/copyparty/pull/703#issuecomment-3228039662

I added a check to make sure the nix packaging doesn't drift apart from the dockerfile.

## Things to package: 

- [x] asmcrypto
- [x] hash-wasm
- [x] markedjs
- [x] DOMPurify (dompf)
- [x] easy-markdown-editor (mde)
- [x] codemirror
- [x] fontawesome
- [x] prism
- [x] sourcecodepro font
- [x] fusepy (in nixpkgs, but the dockerfile patches it)
- [x] [busy-mp3](https://github.com/9001/copyparty/blob/547a7ab1cc7777f3452f441628339850511c8563/scripts/deps-docker/Dockerfile#L65)

This pr should be as complete as it can be right now, would be ideal to have https://github.com/9001/copyparty/pull/868 merged first.


toast003:
The dockerfile does a bunch of patches with sed, so I have chosen to still do them with sed instead of using `substituteInPlace`.

Not best practice (at least in Nixpkgs), but that way we can just copy paste them

9001:
Since you're actually doing this (madlad), I should mention you don't really need all the patches, some are just to reduce the size by removing unnecessary features -- and the main reason the dockerfile is still using Alpine 3.18 to build the deps is because the output from nodejs changed at some point so some of the patches don't apply anymore, though iirc those were not important ones

these can be skipped; just sizegolfing:
* codemirror.patch
* easymde.patch
* marked.patch

these are necessary for scroll-position sync in the markdown editors:
* easymde-ln.patch
* marked-ln.patch

these inline modifications which are made in the dockerfile can be skipped (sizegolfing):
* asmcrypto awk

and the cacophony of seds in the dockerfile for easymde is just to use my modified copy of marked.js instead of having node grab the upstream, there's probably better ways to do this, I"m just banging rocks together here :+1:

toast003:
Thanks for explaining!

I'm running into some weird issue with packaging easymde, I think nix gets confused by having `file:/` urls in package-lock.json?
Anyways I'm not familiar with packaging npm packages with nix so I'll move on to something and try later

toast003:
Ok, should be ready to test now

toast003:
I'm making the assumption that we build the dependencies only if we're building from git and not a pinned release

9001:
cool! if you create a tar.gz of the deps folder I can give it a spin too

toast003:
Sure thing!
[deps.tar.gz](https://github.com/user-attachments/files/23558276/deps.tar.gz)

The `fuse.py` here will probably not work outside my NixOS install cause nix weirdness but yee

9001:
yikes, I completely forgot about this -- sorry!

only remembered because i'm looking at vendoring asmcrypto since it's tricky to build with modern node/npm versions, and went "wait, didn't we have something to build this with nixos...?"

so uhh would now be a good time to merge this, or has it been overtaken?

EDIT: asmcrypto is now vendored btw, bef30ac04d6519c0f30fab6c35c1822c20aa7bdb

toast003:
Hi, I think it would be a good idea to wait a bit before we merge this, since I have an idea for building the web dependencies that would need less maintenance, and I still have to actually see if it would work

toast003:
Basically I recently realized that fixed-output derivations can access the network, so using them we could just extract the `Dockerfile`'s commands into a separate script, and run that (with some slight changes) instead of packaging everything individually.

That would require a lot less Nix code, which is good since copyparty doesn't have that many contributors that want to work or know Nix

Issue #904, opened 2025-10-08
Implications of the Import flag on Subprocess-based hooks


itssoap:
With the new Import flag (-I), which allows the hook to run as a module and not a subprocess/fork, what I am interested to know is, if I have to use this flag with a hook that right now uses `os.chdir()` and `subprocess `to perform certain action, how will I do so? 

A bit detailed explanation/documentation would be appreciated. Like what will be implications for a hook like wget.py. I understand using STDIN and return an RC, but then, what about changing directory and running a command?

Thanks

9001:
`chdir` in an `I` hook is absolutely a bad idea; the problem is that `chdir` (and many other `os` things) affects the whole process -- both of these two example plugins are equally problematic:

```python
import os
def main(a):
    os.chdir("/home/ed/x")
```

```python
import os
import threading
def fun2():
    os.chdir("/home/ed/x")  # even in a thread it affects the whole process
def main(a):
    t=threading.Thread(target=fun2)
    t.start()
    t.join()
```

this will affect some™ parts of copyparty in very unexpected ways... I don't have a complete list of possible side-effects but one example I can think of is `--stackmon` (to periodically dump some debug info to files), which will suddenly start (and then forever continue) writing to another folder after such a hook is run.

`os.chdir` is the most obvious problematic thing I could think of, but there is definitely more -- [another good example](https://github.com/mpv-player/mpv/commit/1e70e82baa9193f6f027338b0fab0f5078971fbe) (not to mention a very entertaining read (CTRL-F `threads` for the relevant part))

and there's things like changing thread affinity, messing with `os.environ`, all has side-effects... so basically, just importing the wrong python module inside an `I` hook is enough to see funky things happen. Sorry for the fuzzy explanation, but the possibilities are just too many and wide-reaching to consider how each would affect which part of copyparty.

However, all that said, just launching a `subprocess` *should* be fine! And `subprocess.run` has the `cwd` option to launch the process with another PWD which should be a good (and safe) alternative to `os.chdir` -- just don't use the `preexec_fn` property which is offered by for example `suprocess.Popen` (that one has pretty big warnings in the python docs too)

9001:
specifically for an import-safe version of wget.py, I think it could be a good example, so I'm making `wget-i.py` now, sec

itssoap:
[...]

PR #908, opened 2025-10-09
Add an import-safe yt-dlp hook (#904)

itssoap:
Based on the conversation in #904 and the wget.py hook


itssoap (commented):
My part is done here

PR #914, opened 2025-10-11
fix: ftp append

Audionut:
When this is received:

`{time} {IP}   [USER] APPE b'\\\\?\\{file_location}' completed=1 bytes=1892120 seconds=0.172`

copyparty currently unlinks, and overwrites the existing file, meaning live updates to the file are not correctly appended.

copyparty only uses `ftp_STOR` with `mode=w` by default.
pyftpdlib's default APPE implementation calls [ftp_STOR(file, "a")] automatically.

This PR adds an `a` mode condition for the unlink handling to prevent this behavior, and allow appending data to the existing file.

### DIFF
--- a/copyparty/ftpd.py
+++ b/copyparty/ftpd.py
@@ -249,7 +249,8 @@ class FtpFs(AbstractedFS):
                 need_unlink = False
                 td = 0
 
-        if w and need_unlink:
+        # Don't unlink file for append mode
+        if w and need_unlink and "a" not in mode:
             if td >= -1 and td <= self.args.ftp_wt:
                 # within permitted timeframe; unlink and accept
                 do_it = True
--- a/copyparty/ftpd.py
+++ b/copyparty/ftpd.py
@@ -250,9 +250,9 @@ class FtpFs(AbstractedFS):
                 td = 0
 
         # Don't unlink file for append mode
-        if w and need_unlink and "a" not in mode:
+        if w and need_unlink:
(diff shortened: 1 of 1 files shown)


9001:
Thanks! But while this does fix the bug, there's one issue left to solve --

Copyparty in general does not allow for modifying files after they've been uploaded; this is for example to prevent someone with write-only access from corrupting existing files by appending garbage to them. The `ftp-wt` option is the grace-period in which an aborted transfer is allowed to be resumed, so I think the following approach would be better:

```patch
@@ -252,5 +252,5 @@ class FtpFs(AbstractedFS):
         if w and need_unlink:
             if td >= -1 and td <= self.args.ftp_wt:
-                # within permitted timeframe; unlink and accept
+                # within permitted timeframe; allow overwrite or resume
                 do_it = True
             elif self.args.no_del or self.args.ftp_no_ow:
@@ -268,6 +268,6 @@ class FtpFs(AbstractedFS):
             if not do_it:
                 raise FSE("File already exists")
-
-            wunlink(self.log, ap, VF_CAREFUL)
+            elif "a" not in mode:
+                wunlink(self.log, ap, VF_CAREFUL)
 
         ret = open(fsenc(ap), mode, self.args.iobuf)
```

if this looks good, could you please give it a spin?

Audionut:
That works for my use-case, thanks.

9001:
sweet! will be in the next ver :>

PR #916, opened 2025-10-11
view .cbz in browser

Scotsguy:
demo:
https://github.com/user-attachments/assets/ed19f663-055d-4e52-9c76-6ac3d8f96db5

adds functionality to allow browsing .cbz directly in the browser, without downloading them and using a separate program. meant for quickly inspecting the contents, less so for reading

adds two new api calls, ?zls and ?zget, which return a file listing of a zip file and a specific file in the archive, respectively.
uses the zipfile module, so no support for .cbr etc

i ended up refactoring how bbox is initialized a bit, which hopefully didn't break anything?

also a somewhat-related fix for the error message when thumbnailing empty .cbz

### DIFF
--- a/copyparty/httpcli.py
+++ b/copyparty/httpcli.py
@@ -1554,10 +1554,12 @@ class HttpCli(object):
         try:
             with zipfile.ZipFile(abspath, "r") as zf:
                 zi = zf.getinfo(inner_path)
+                if zi.file_size >= maxsz:
+                    raise Pebkac(404, "zip bomb defused")
                 with zf.open(zi, "r") as fi:
                     self.send_headers(length=zi.file_size, mime=guess_mime(inner_path))
 
-                    remains = sendfile_py(
+                    sendfile_py(
                         self.log, 0, zi.file_size,
                                 fi,
                                 self.s,
@@ -1567,20 +1569,6 @@ class HttpCli(object):
                                 {},
                                 "",
                                 )
-                    # fd, ret = tempfile.mkstemp("." + inner_path.rsplit(".", 1)[0])
-                    # fsz = 0
-                    # with os.fdopen(fd, "wb") as fo:
-                    #
-                    #     while True:
-                    #         buf = fi.read(32768)
-                    #         if not buf:
-                    #             break
-                    #
-                    #         fsz += len(buf)
-                    #         if fsz > maxsz:
-                    #             raise Exception("zipbomb defused")
-                    #
-                    #         fo.write(buf)
         except KeyError:
             raise Pebkac(404, "no such file in archive")
         except (zipfile.BadZipfile, RuntimeError):
--- a/copyparty/httpcli.py
+++ b/copyparty/httpcli.py
@@ -1560,20 +1560,21 @@ class HttpCli(object):
                     self.send_headers(length=zi.file_size, mime=guess_mime(inner_path))
 
                     sendfile_py(
-                        self.log, 0, zi.file_size,
-                                fi,
-                                self.s,
-                                self.args.s_wr_sz,
-                                self.args.s_wr_slp,
-                                not self.args.no_poll,
-                                {},
-                                "",
-                                )
+                        self.log,
+                        0,
+                        zi.file_size,
+                        fi,
+                        self.s,
+                        self.args.s_wr_sz,
+                        self.args.s_wr_slp,
+                        not self.args.no_poll,
+                        {},
+                        "",
+                    )
         except KeyError:
             raise Pebkac(404, "no such file in archive")
         except (zipfile.BadZipfile, RuntimeError):
             raise Pebkac(404, "requested file is not a valid zip file")
-
 
     def handle_propfind(self) -> bool:
         if self.do_log:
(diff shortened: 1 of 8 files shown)


copyparty/httpcli.py
```diff
@@ -6576,6 +6638,7 @@ def tx_browser(self) -> bool:
             "taglist": [],
             "have_tags_idx": int(e2t),
             "have_b_u": (self.can_write and self.uparam.get("b") == "u"),
+            "have_zls": int(not self.args.no_zls),
```
9001: since this is a static value defined by the config, it can be added to `js_htm` in authsrv instead to have it automatically added to to the window/global scope on page load

copyparty/web/browser.html
```diff
@@ -139,7 +139,8 @@ <h2 id="wfp"><a href="{{ r }}/?h" id="goh">control-panel</a></h2>
 			have_tags_idx = {{ have_tags_idx }},
 			sb_lg = "{{ sb_lg }}",
 			logues = {{ logues|tojson if sb_lg else "[]" }},
-			ls0 = {{ ls0|tojson }};
+			ls0 = {{ ls0|tojson }},
+			have_zls = {{ have_zls }};
```
9001: by moving have_zls to js_htm in authsrv, this won't be necessary  :>

docs/devnotes.md
```diff
@@ -199,6 +199,8 @@ authenticate using header `Cookie: cppwd=foo` or url param `&pw=foo`
 | GET | `?th` | get image/video at URL as thumbnail |
 | GET | `?th=opus` | convert audio file to 128kbps opus |
 | GET | `?th=caf` | ...in the iOS-proprietary container |
+| GET | `?zls` | get listing of filepaths in zip file at URL |
+| GET  | `?zget=path` | get specific file from inside a zip file at URL |
```
9001: extra whitespace (because I'm not finding anything else to complain about :grin: )

9001:
awesome stuff! only finding some tiny nitpicks on the python side of things, good work :>

as for the javascript, turns out the arrow syntax killed support for older browsers :stuck_out_tongue_closed_eyes:  here's a (very) quick fix for that unless you have a better idea:

```patch
@@ -199,3 +199,3 @@
                 e.preventDefault ? e.preventDefault() : e.returnValue = false;
-                fillCbzGallery(gallery, cbzElement, eventHandler).then(() => {
+                fillCbzGallery(gallery, cbzElement, eventHandler).then(function() {
                     prepareOverlay(gallery, userOptions);
@@ -203,3 +203,3 @@
                     }
-                ).catch((reason) => {
+                ).catch(function(reason) {
                     console.error("cbz-ded", reason);
@@ -229,3 +229,3 @@
         return fetch(zlsHref)
-            .then(response => {
+            .then(function(response) {
                 if (response.ok) {
@@ -236,7 +236,7 @@
             })
-            .then((fileList) => {
-                var imagesList = fileList.filter((name) =>
-                    name.indexOf(".") !== -1
+            .then(function(fileList) {
+                var imagesList = fileList.filter(function(name) {
+                    return name.indexOf(".") !== -1
                     && cbz_pics.indexOf(name.split(".").pop()) !== -1
-                ).sort();
+                }).sort();
 
@@ -246,3 +246,3 @@
 
-               imagesList.forEach((imageName, index) => {
+               imagesList.forEach(function(imageName, index) {
                     var imageHref = href
```

of course the new feature itself fails in ie11 because it doesn't have fetch, but I am NOT AT ALL concerned about that :grin:

9001:
come to think of it, `?zls` currently returns the list of filenames, which for this usecase is all we need -- but a list of `{fn:"filename", sz:size, ts:timestamp}` could be neat for other usecases... Though I don't think it's important to consider, we could instead add it as an option later if it's ever necessary (`?zls=i` for info or something)

Scotsguy:
> awesome stuff! only finding some tiny nitpicks on the python side of things, good work :>
> 
> as for the javascript, turns out the arrow syntax killed support for older browsers 😝 here's a (very) quick fix for that unless you have a better idea:

ah, i was thinking it was fine because fetch doesn't work anyway, but didn't think of it causing syntax errors, whoops

> come to think of it, `?zls` currently returns the list of filenames, which for this usecase is all we need -- but a list of `{fn:"filename", sz:size, ts:timestamp}` could be neat for other usecases... Though I don't think it's important to consider, we could instead add it as an option later if it's ever necessary (`?zls=i` for info or something)

I'll turn it into a dictionary but without adding any other properties, that way we can add other properties without breaking existing code

9001:
just to clarify, i'm thinking we should keep the outer container that's returned a list, so it'll have all the files in the same order as the zip has them -- `return [{"fn":"a.jpg"}, {"fn":"b.jpg"}]`

copyparty/httpcli.py
Scotsguy: fixed. `have_tags_idx` looked too similar, but i suppose that's a volume flag so not static

9001:
sweet, thx again :grin: :+1:

Scotsguy:
thanks for merging :tada:

PR #921, opened 2025-10-14
Add the ability for on message hooks to see selected files

carson-coder:

This pull request makes the browser send another param `sel` to the server when sending a message. This param is parsed and passed to any `xm` hooks. This allows on_message hooks to see selected files. This also adds a param to runhook that allows code that runs hooks to pass arbitrary values into the hook.

### DIFF
--- a/copyparty/httpcli.py
+++ b/copyparty/httpcli.py
@@ -2146,8 +2146,18 @@ class HttpCli(object):
                     try:
                         zb = unquote(buf.replace(b"+", b" "))
                         plain = zb.decode("utf-8", "replace")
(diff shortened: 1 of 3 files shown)


carson-coder:
https://git.carsoncoder.com/carsoncoder/CopypartyServerHooks/src/branch/main/on_message/ffmpeg_hook.py

Here is a example

9001:
Thanks -- I went about it slightly differently to avoid any possible edgecases in url-decoding, the `body` field can be used for more complex situations, but splitting the `txt` on `\n` should be good enough for most cases

PR #936, opened 2025-10-20
[Esperanto TL] fix capitalization on filesize option

slashdevslashurandom:
Okay, now that each translation's strings are in their own files, doing small updates and such should be much easier, right?

(this is a quick web-edit, I hope it didn't cause any damaging changes to formatting or something, because other projects tend to frown on that. I can do a proper edit, if necessary)

(P.S. also the machine-translated lines warning was edited a bit)


9001:
thanks!

PR #962, opened 2025-10-31
nixos: Only create and bind volumes without variables

nicomem:
copyparty allows one to define dynamic volumes using variables, for example: a `/u/${u}` storage with the path `/storage/u/${u}`.

However, doing that with the NixOS module results in errors as the modules tries to create and bind the folder `/storage/u/${u}`, which is not what is wanted.

This PR fixes that by filtering out the volumes that contain variables in their path (i.e. contains `${`) from folders creation and binding in the service.

For users that defines a parent volume (e.g. a `/u` volume with the `/storage/u` path), they do not have to do anything as:
- `/storage/u` is created from the module config and `/storage/u/${u}` can be created dynamically by copyparty
- `/storage/u` is bound in the service, so `/storage/u/${u}` can be accessed and written

### DIFF
--- a/contrib/nixos/modules/copyparty.nix
+++ b/contrib/nixos/modules/copyparty.nix
@@ -48,6 +48,8 @@ let
 
   accountsWithPlaceholders = mapAttrs (name: attrs: passwordPlaceholder name);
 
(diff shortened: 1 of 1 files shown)


toast003 (approved):

9001:
thanks!

PR #963, opened 2025-10-31
nixos: Use volume/global chmod-d for module directory creation

nicomem:
Currently, the NixOS module automatically creates all the volumes' directories and sets their permissions to `755`.

This PR makes changes the permissions that is set to take into account the volume and/or global settings:
- If the volume's `chmod_d` setting is specified, use this for the mode
- Else, if the global `chmod-d` setting is specified, use this for the mode
- Else, use the previous default value of `755`

In all cases, the `:` prefix is used, so the tmpfile rule will not change the mode if the directory already exists (which follows the behaviour of `chmod-d` and `chmod_d` which only apply for new directories).


toast003 (approved):

9001:
thanks!

PR #965, opened 2025-11-01
focus up2k when dragging from browser

Scotsguy:
otherwise the toast tells you to drag onto a button that may not be visible

### DIFF
--- a/copyparty/web/up2k.js
+++ b/copyparty/web/up2k.js
@@ -1065,8 +1065,11 @@ function up2k_init(subtle) {
                         if (toast.txt == L.u_uri)
                             toast.hide();
                     }
-                    else
+                    else {
+                        if (!QS('#op_up2k.act'))
+                            goto('up2k');
                         return toast.inf(10, L.u_uri) || true;
+                    }
                 }
 
             if (!ok)


9001:
so this has a tiny bit of backstory to it -- the reason it requires the button as drop-target is mainly to help a guy with tremors, and I'm concerned he wouldn't much enjoy the sudden layout-shift of switching to the upload-UI -- though I also agree it's a huge inconvenience for everyone intentionally trying to do this ...

but I think the following should work, it would watermark all drags coming from the same tab  (for example by accidentally dragging a thumbnail a tiny bit) and it should only then trigger the button requirement,

this patch is ontop of hovudstraum, not based on the existing changeset,

```patch
diff --git a/copyparty/web/up2k.js b/copyparty/web/up2k.js
index d4dd375d..d422a65c 100644
--- a/copyparty/web/up2k.js
+++ b/copyparty/web/up2k.js
@@ -798,4 +798,5 @@ function up2k_init(subtle) {
         "gotallfiles": [gotallfiles]  // hooks
     };
+    r.drag_id = "application/x-" + r.tact;
 
     setTimeout(function () {
@@ -1054,4 +1055,10 @@ function up2k_init(subtle) {
         return onovercmn(this, e, true);
     }
+    document.body.ondragstart = function (e) {
+        try {
+            e.dataTransfer.setData(r.drag_id, "x");
+        }
+        catch (ex) { }
+    };
     function onovercmn(self, e, btn) {
         try {
@@ -1061,5 +1068,5 @@ function up2k_init(subtle) {
                     ok = true;
                 else if (dt[a] == 'text/uri-list') {
-                    if (btn) {
+                    if (btn || !has(dt, r.drag_id)) {
                         ok = true;
                         if (toast.txt == L.u_uri)
```

could you take this for a spin for a while? if it seems to work, let's go with this :+1:

Scotsguy:
ah, i didn't know that had accessibility reasons behind it, good to know

a summary of behaviour with and without your patch (tested in firefox):
| |hovudstraum|patch|
|-|-|-|
|d&d file from downloads toolbar|"use button" toast|fullscreen up2k d&d ui|
|d&d image from other window|"use button" toast, uploads 0 files| fullscreen up2k d&d ui, uploads 0 files|
|d&d image from other copyparty window|"use button" toast on sender and receiver, uploads 0 files|up2k ui on receiver, toast on sender, uploads 0 files|
|d&d grid thumbnail|"use button" toast, uploads 0 files|"use button" toast, uploads 0 files|
|d&d link from other window/browser|"use button" toast, uploads 0 files|fullscreen up2k d&d ui, uploads 0 files|
|drag text from other window/browser|nothing happens|nothing happens|
|drag image from chromium into firefox|navigates to image link|navigates to image link|

i think a few of these could be improved, namely:
- the toast is superfluous since it doesn't show up when trying to drag from another window
- d&d-ing images doesn't actually work, only files

apart from that, I think the patch works well

(i also noticed that you can get the fullscreen up2k ui stuck by cancelling the drag with escape, you have to click to get it unstuck. probably need to subscribe to ondragend or something)

9001:
ah right, yeah Firefox doesn't support it i think, just chrome, as unfortunate as that is ;_;

and good catch with the UI getting stuck -- if you get bored I'll merge what you cook up, otherwise I'll be back in action around last weekend this month 👍

Scotsguy:
`dom.events.datatransfer.imageasfile.enabled` makes images drag like images (but not as links anymore)
https://bugzilla.mozilla.org/show_bug.cgi?id=1437126

I'll see if i can spare some time to sand the edges off drag&drop, then

Scotsguy:
closing to clean up

PR #977, opened 2025-11-04
Fix broken links in the control panel

Habetdin:
While checking `my.copyparty.host/some-subdir/?h` control panel, I noticed an active upload, yet was still to realize that the upload directory link is relative to that `some-subdir`. When I clicked the link, all I got was a 404 error.

This PR fixes active download/upload links in the control panel.


9001:
nice, thx!

PR #981, opened 2025-11-04
Add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

there might still be a few false positives to catch since overall a good number of fixes


.github/workflows/codespell.yml
```diff
@@ -0,0 +1,25 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
+  push:
+    branches: [hovudstraum]
+  pull_request:
+    branches: [hovudstraum]
```
yarikoptic, line 9: or should apply to all?

copyparty/web/browser.js
```diff
@@ -4142,7 +4142,7 @@ var fileman = (function () {
 		var ire = ebi('rn_re'),
 			ifmt = ebi('rn_fmt'),
 			ipre = ebi('rn_pre'),
-			idel = ebi('rn_pdel'),
```
Scotsguy: false positive

copyparty/cert.py
```diff
@@ -23,7 +23,7 @@ def ensure_cert(log: "RootLogger", args) -> None:
     """
     the default cert (and the entire TLS support) is only here to enable the
     crypto.subtle javascript API, which is necessary due to the webkit guys
-    being massive memers (https://www.chromium.org/blink/webcrypto)
```
Scotsguy: false positive

docs/versus.md
```diff
@@ -259,7 +259,7 @@ symbol legend,
   * you can successfully play `$'\355\221'` with mpv through mounting a remote copyparty server with rclone, pog
 * `a`/copyparty remarks:
   * extremely minimal samba/cifs server
-  * netscape 4 / ie6 support is mostly listed as a joke altho some people have actually found it useful ([ie4 tho](https://user-images.githubusercontent.com/241032/118192791-fb31fe00-b446-11eb-9647-898ea8efc1f7.png))
```
Scotsguy: altho -> although but not tho -> though?

docs/changelog.md
Scotsguy, line 1: why change the changelog retroactively?

Scotsguy (changes requested):
A bunch of these are imo differences of style and not typos, some false positives too

copyparty/cert.py
yarikoptic: that's cool -- I didn't even know about such modern word ;) excluding!

copyparty/web/browser.js
yarikoptic: excluding

docs/changelog.md
yarikoptic: it is still a part of the "documentation" and I do not see why to keep it "buggy". But it is easy to skip if you insist.

docs/versus.md
yarikoptic: ;-) nice observation.  up to you -- we could exclude `altho` for symmetry or I could suggest `tho -> though` to codespell and fix here. Your desire?

copyparty/cert.py
yarikoptic: I force pushed with this excluded... not sure why here doesn't say "outdated".
image

docs/changelog.md
Scotsguy: It's also I think copied from https://github.com/9001/copyparty/releases for each release and therefore would get out of sync with those. it's probably not that big of a deal, tho

docs/versus.md
Scotsguy: it's not my decision to make in the end, either :) I'm just some guy, you know
(I like the casual tone of altho, if it matters)

PR #1008, opened 2025-11-16
Allow reading passwords from enviroment variable or file

### LINKED ISSUE #1007: Read passwords from files (at least on docker)
**is your feature request related to a problem? Please describe.**
Reading secrets from config files is not exactly a good idea. While it may be better than just hardcoding it, it's just barely above it.

A config file should have options to read passwords from a file that the process has access to.

**Describe any alternatives you've considered**
I have considered environment variables. While they are safer than config files, they are just barely. They also in docker compose, **have** to be hardcoded. I think they should be an option but not the only one.

ltecheroffical:
Allows reading passwords from an environment variable or file with the @ character at the beginning of the file. A password URL (What i'm calling this) can be created with this format `@url_type:url_location`.

There are two URLs in this PR already implemented:
- `env`: Uses the url location as the environment variable to read to get the password
- `file`: Uses the url location as the path to the file to read to get the password

### Breaking Change
If a preexisting password has `\` or `@` as their first character, the `\` or `@` will be ignored. This can be fixed by using a `\` as the first character before their password which does not include the first `\` in the final password.

Resolves #1007

### DIFF
--- a/copyparty/authsrv.py
+++ b/copyparty/authsrv.py
@@ -1718,6 +1718,33 @@ class AuthSrv(object):
         flags[name] = vals
         self._e("volflag [{}] += {}  ({})".format(name, vals, desc))
 
+    def _parse_password(self, password: str) -> str:
+        """
+        Parses a config file password into a real password
+        """
+        SPECIAL_PASSWORD_CHARACTER = "@"
+        ESCAPE_CHARACTER = "\\"
+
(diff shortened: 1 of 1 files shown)


ltecheroffical:
Oops, forgot to sign 😅

ltecheroffical:
Ok, fixed signing

PR #1029, opened 2025-11-22
fix: front-end validation for password field length

grantbacon:
Adds max length validation to password field, since its limited by 64 chars in backend.

Considered maxlength, but this shows more a useful validation error instead of truncating the field.

### DIFF
--- a/copyparty/web/splash.html
+++ b/copyparty/web/splash.html
@@ -111,7 +111,7 @@
 		<div>
 			<form id="lf" method="post" enctype="multipart/form-data" action="{{ r }}/{{ qvpath }}">
 				<input type="hidden" id="la" name="act" value="login" />
-				<input type="password" id="lp" name="cppwd" placeholder=" password" maxlength="64" />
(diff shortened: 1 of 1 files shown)


copyparty/web/splash.html
```diff
@@ -111,7 +111,7 @@ <h1 id="z">unlock this share:</h1>
 		<div>
 			<form id="lf" method="post" enctype="multipart/form-data" action="{{ r }}/{{ qvpath }}">
 				<input type="hidden" id="la" name="act" value="login" />
-				<input type="password" id="lp" name="cppwd" placeholder=" password" maxlength="64" />
+				<input type="password" id="lp" name="cppwd" placeholder=" password" oninput="this.setCustomValidity('')" oninvalid="this.setCustomValidity('Password must be between 1 and 64 characters')" "pattern="^.{1,64}$" />
```
inexpeditus: `pattern="^.{1,64}$"`

grantbacon: oops! fixed :) thanks

9001:
thanks for the idea! was able to reuse the machinery added in #1105 to have it display the warning immediately as the field gets too long, while also working on older browsers :>

PR #1031, opened 2025-11-23
feat: WIP Rebindable hotkeys

grifinas:
This is just me telegraphing upcoming PR and trying to get early(ish) feedback

Motivation:
* As a user I'd like to be able to rebind the hotkeys to something i find more intuitive
* As a user with accessibility needs, I'd like to rebind the hotkeys to keys that better suit my abilities.
* As a developer, I want to add new bindings without risking conflicts with existing user workflows

Description:
This PR adds the ability for the user to rebind existing hotkeys. It refactors all existing hotkeys behind actions. Allows new developers to, at a glance see what actions are available and to compose/modify them easier.

Notes:
* UI's not done. Need buttons to rebind, delete, add, reset keybind. Also will change how HTML is constructed
* Keybinding structs are a bit of a mess, will eventually expose a more cohesive thing
* Added lazyLoad JS function to allow devs to add scripts without changing HTML files. Will probably not end up using it in final PR, since it complicates the dependency dynamics
* Not tested on Internet Explorer, i don't have Windows. 
    * Also there's no contribution guidelines for browser support. https://github.com/9001/copyparty?tab=readme-ov-file#browser-support Implies that ie6 is a target, but not all features are supported by it
    * On the other hand, https://github.com/9001/copyparty/blob/hovudstraum/CONTRIBUTING.md makes no mention of targeted versions. Would be nice to clear that up.
    * Personally, I would like to not support browsers that https://developer.mozilla.org does not even list in compatibility charts. 
* Assuming Older browser support is optional I would at least like to use templates instead of generating string HTML
* Need to go through all other UI elements where hotkeys are hinted and make them resolve the hotkeys that the user is currently using
* Need to make ? hotkey also render the new hotkeys, not the OG
* Eventually, though maybe not with this iteration, i'd be nice if adding a hotkey was just the modification of one struct with maybe a function definition for the implementation. Devs could refer the the action when they want to execute it making the code vastly more readable. e.g. `execute(HOTKEY_ACTIONS.PLAY_PAUSE)`


NecRaul:
Not sure if you're still planning to continue this, but it would be great to see it finished since customizable hotkeys would be really useful.

Also, if you end up expanding it further, it could be nice to support optional preset keybinding schemes (e.g. vi or emacs) for users who already have muscle memory for those.

PR #1037, opened 2025-11-25
Video playback speed setting in bbox

SelfishPig:
Added a playback speed button in the bbox overlay.


9001:
should explain why i've been delaying this -- it necessitates a new layout of the toolbar, since it is now finally too wide to fit on phones. I will merge this eventually but I need to find the time, sorry

Issue #1047, opened 2025-11-28
RSS Feed - Use tag for Title if available


djjeane:
**Desire**
When generating the RSS body, if a title tag is available, use that tag value for the title field of the xml sent. 

Currently, the filename is always used for the title, even when a title tag is present. Its a bit annoying to have the full filename displayed when using podcast readers. If the file is properly tagged, we can source a clean name from the file tags.

9001:
could this possibly cause issues for some readers if there are multiple files with the same title? I assume the answer is no, but I'm not very familiar with rss -- and I guess this should be configurable either way, in case someone wants something like "artist - title" and not just the title too

djjeane:
I dont expect RSS readers would complain with duplicate titles, as long as the GUID is unique. But you are right, its probably best if its configurable. I think filename is a fine default for title, but people with well tagged files will appreciate the support when streaming files via RSS.

PR #1053, opened 2025-11-29
Enhanced tag support for RRS podcast steam

djjeane:
To better support the podcast rss feed

- Uses Date tag to set published data of item in steam
- Sources Title from title tag if present
- Restructures description from tags if present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1067,6 +1067,12 @@ url parameters:
   * `n` = filename
   * `a` = filesize
   * uppercase = reverse-sort; `M` = oldest file first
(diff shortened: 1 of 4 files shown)


djjeane:
Took a stab at the implementation of this issue. Let me know if this needs any changes before the merge.

#1047

9001:
thanks for the contribution! I used as inspiration for a slightly different take; should be pretty flexible now, and performance has now only been reduced by 4% compared to before these changes. This will be in the next release, probably in about 20 hours.

PR #1054, opened 2025-11-29
fix: round bad floating point arithmetic

shermanhlc:
when using the scroll wheel on the volume bar, it adjusts in factors of 2, except when going from 92 -> 89, due to bad floating point arithmetic (0.94 - 0.02 is calculated as 91.9...9)

https://github.com/user-attachments/assets/d0d09634-14ba-4e86-b473-0eff513756c7


9001:
keen eye! thanks o/

PR #1097, opened 2025-12-11
fix blank password handling/ui when banning

stackxp:
Various fixes, including:
- blank passwords are now eligable for banning
- blank passwords now show `msg.html`, as if entering a wrong password
- `msg.html` initially wouldn't display properly right before getting banned, fixed by embedding the css

I hope that blank passwords are not some kind of exception to regular wrong passwords, regarding the banning and ui behavior.


stackxp:
Also, please ignore the random change at the end of util.py, I created this PR while a new version was pushed to `hovudstraum` and it broke some things.

PR #1105, opened 2025-12-14
add blank password handling/fix ban behavior

stackxp:
- Adds a check to the control panel/login screen that prevents sending an empty password.
- Also fixes a bug with the css of the message screen right before getting banned.
- Adds two new translation keys:
  - nop: "please enter a password"
  - nou: "please enter a username and password"


9001:
thanks! just made some slight tweaks (mostly restoring support for custom fonts)

PR #1120, opened 2025-12-17
Parse total space from `vmaxb` volflag

rabid-dev:
Added a couple lines to return the total space available in a volume based on the `vmaxb` volflag, and change how the used/free space is calculated when using that.

Example:
I have a ~600GB HDD. I have created 3 volumes, dir1, dir2, dir3.
Dir1 has **100GB** `vmaxb` volflag, Dir2 has **50GB** `vmaxb` volflag, and Dir3 has no `vmaxb` volflag set.

Without the changes of this PR (this is how currently works) (sorry for the spanish language) my drives look like this:

unlimited

With the changes of this PR, you can see windows shows correctly the quota:

limited

I have not tested it a lot, just manually with some volumes I have configured in the pictures, but I can try to add some unit tests if you want (although I might need some help with that...)

====

<details>
<summary>disclaimer after reading `contributing.md` (click to expand)</summary>
I used an AI to ask it 2 things:

1) where was the code that returned the total space available and the total space used in order to insert these changes

2) help to calculate the space used on the volume (to be precise, the method call `self.conn.hsrv.broker.ask("up2k.get_volsizes", [vn.realpath]).get()[0]` from line 1838)

I did this because I just wanted to see if I could implement it but I didn't want to invest a lot of time as I have other projects that require my attention (but my ADHD brain wouldn't let me sleep over having the network drives with the wrong available/used space values...)

I've set this disclaimer because as I was opening this PR i saw the note in the contributing guidelines to not use AI, and I'm truly sorry as I was not aware of this when I was doing the changes.

I will understand if you decide to decline the pull request, and i will not get angry/annoyed at the decision. You've made a truly great software and if you decide to keep it fully organic with no AI at all then I am not going to complain.
</details>

====

I think this PR complies with the DCO anyways (or I hope it still does); https://developercertificate.org/

### DIFF
--- a/copyparty/httpcli.py
+++ b/copyparty/httpcli.py
@@ -117,7 +117,6 @@ from .util import (
     undot,
     unescape_cookie,
     unquotep,
-    unhumanize,
     vjoin,
     vol_san,
     vroots,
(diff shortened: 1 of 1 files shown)


9001:
thanks! I like the idea and want to merge this before the next release, possibly with some tiny changes -- but it might be a few weeks :+1:

rabid-dev:
> thanks! I like the idea and want to merge this before the next release, possibly with some tiny changes

yay! 🎉
> -- but it might be a few weeks 👍

no problem at all! i even consider that to be quite fast in fact, due to having christmas and new year coming soon 👀

9001:
alright, adjusted it to reply with the available disk space when that is smaller than the specified vmaxb

thx again o/

rabid-dev:
Hi again @9001 ! Thanks for taking the time to fix that issue, I didn't take into account the disk space being smaller than the quota 👀 

However, with this new version you don't seem to update the `btot` variable, only the `bfree` one... I just tested this and it shows the following (dir1 had 100GB quota on a 585GB hard drive)
imagen

Maybe below line 1841 there should be added this line?
`btot = min(btot, vn.lim.vbmax)`

9001:
good point :grin:

rabid-dev:
> good point 😁

LGTM now!
imagen

PR #1128, opened 2025-12-19
up2k: add "skip if file exists"

Scotsguy:
new option for the up2k overwrite switch: if the server has a file with the same name as is being currently uploaded, don't upload it

this one was missing for me because it's the only option to skip uploading, all others upload it in one place or another

### DIFF
--- a/copyparty/httpcli.py
+++ b/copyparty/httpcli.py
@@ -2875,7 +2875,7 @@ class HttpCli(object):
         body["addr"] = self.ip
         body["vcfg"] = dbv.flags
 
-        if not self.can_delete:
+        if not self.can_delete and not body.get("replace") == "skip":
             body.pop("replace", None)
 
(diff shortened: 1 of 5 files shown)


9001:
thanks! I plan to merge this before the next release, but it might be a few weeks :+1:

dineshgyl:
but what if file contents are changed? will it still skip upload?

9001:
> but what if file contents are changed? will it still skip upload?

of course it will; that is the purpose of this new mode.

if you do not want that behavior then use the other existing modes of that button

Scotsguy:
> but what if file contents are changed? will it still skip upload?

if the file you're uploading to has the same contents (and it's indexed) it will skip the upload regardless of overwrite settings by comparing the hash. the overwrite settings only apply if they're different

PR #1135, opened 2025-12-23
Add right-click menu

stackxp:
This adds a right-click menu to copyparty allowing you to:
- Open files in different ways (Open in MD editor, Play video as audio, etc.)
- Cut, Copy, Paste and Delete stuff
- Download files normally or as an archive
- mess with selections

Apart from that, this also:
- Adds a new section to the options page
- Allows you to use the regular right-click menu when double right-clicking
- Adds 19 new translation keys

I didn't implement the grid view yet but I hope that isn't a problem :)

### DIFF
--- a/copyparty/web/browser.js
+++ b/copyparty/web/browser.js
@@ -9622,8 +9622,8 @@ var rcm = (function () {
 		clmod(ebi('rpst'), 'hide', !has_clip);
 		clmod(ebi('rs3'), 'hide', !has_sel || !has_clip);
 
-		menu.style.left = x + 1 + 'px';
-		menu.style.top = y + 1 + 'px';
+		menu.style.left = x + 5 + 'px';
+		menu.style.top = y + 5 + 'px';
 		menu.style.display = 'block';
 		menu.focus();
 	}
@@ -9638,14 +9638,17 @@ var rcm = (function () {
 		menu.style.display = '';
 	}
 
+	var lrcTime = null;
 	ebi('wrap').oncontextmenu = function(e) {
 		hide(true);
-		console.log(e.detail)
-		if (thegrid.en || !r.enabled || (r.double && e.detail > 1))
+		if (thegrid.en || !r.enabled || (r.double && lrcTime && e.timeStamp - lrcTime < 500)) {
 			return true;
-		ev(e);
-		show(e.x, e.y, e.target);
-		return false;
+		} else {
+			lrcTime = e.timeStamp;
+			ev(e);
+			show(e.clientX, e.clientY, e.target);
+			return false;
+		}
 	};
 	menu.onblur = function() {setTimeout(hide)};
 
--- a/copyparty/web/browser.js
+++ b/copyparty/web/browser.js
@@ -9631,7 +9631,7 @@ var rcm = (function () {
 
 	ebi('wrap').oncontextmenu = function(e) {
 		hide(true);
-		if (thegrid.en || !r.enabled || e.shiftKey)
+		if (thegrid.en || !r.enabled || e.shiftKey || menu.style.display)
 			return true;
 		else {
 			ev(e);
--- a/copyparty/web/browser.js
+++ b/copyparty/web/browser.js
@@ -223,6 +223,7 @@ if (1)
 		"cl_reset": "reset",
 		"cl_hpick": "tap on column headers to hide in the table below",
 		"cl_hcancel": "column hiding aborted",
+		"cl_rcm": "right-click menu",
 
 		"ct_grid": '田 the grid',
 		"ct_ttips": '◔ ◡ ◔">ℹ️ tooltips',
@@ -265,6 +266,8 @@ if (1)
 		"cdt_lim": "max number of files to show in a folder",
 		"cdt_ask": "when scrolling to the bottom,$Ninstead of loading more files,$Nask what to do",
 		"cdt_hsort": "how many sorting rules (&lt;code&gt;,sorthref&lt;/code&gt;) to include in media-URLs. Setting this to 0 will also ignore sorting-rules included in media links when clicking them",
+		"cdt_ren": "enable custom right-click menu",
+		"cdt_rno": "enable normal menu when double right-clicking",
 
 		"tt_entree": "show navpane (directory tree sidebar)$NHotkey: B",
 		"tt_detree": "show breadcrumbs$NHotkey: B",
@@ -640,6 +643,22 @@ if (1)
 		"ur_um": "Finished;\n{0} uploads OK,\n{1} uploads failed, sorry",
 		"ur_sm": "Finished;\n{0} files found on server,\n{1} files NOT found on server",
 
+		"rc_opn": "Open",
+		"rc_ply": "Play",
+		"rc_pla": "Play as audio",
+		"rc_txt": "Open in text editor",
+		"rc_md": "Open in markdown editor",
+		"rc_dl": "Download",
+		"rc_zip": "Download as archive",
+		"rc_del": "Delete",
+		"rc_cut": "Cut",
+		"rc_cpy": "Copy",
+		"rc_pst": "Paste",
+		"rc_nfo": "New folder",
+		"rc_nfi": "New file",
+		"rc_sal": "Select all",
+		"rc_sin": "Invert selection",
+
 		"lang_set": "refresh to make the change take effect?",
 };
 
@@ -966,6 +985,13 @@ ebi('op_cfg').innerHTML = (
 	'	</div>\n' +
 	'</div>\n' +
 	'<div><h3>' + L.cl_keytype + '</h3><div><select id="key_notation"></select></div></div>\n' +
+	'<div>' +
+	'	<h3>' + L.cl_rcm + '</h3>' +
+	'	<div>' +
+	'		<a id="ren" class="tgl btn" href="#" tt="' + L.cdt_ren + '">enable</a>' +
+	'		<a id="rno" class="tgl btn" href="#" tt="' + L.cdt_rno + '">double-click</a>' +
+	'	</div>' +
+	'</div>' +
 	'<div><h3>' + L.cl_hiddenc + ' &nbsp;' + (MOBILE ? '<a href="#" id="hcolsh">' + L.cl_hidec + '</a> / ' : '') + '<a href="#" id="hcolsr">' + L.cl_reset + '</a></h3><div id="hcols"></div></div>'
 );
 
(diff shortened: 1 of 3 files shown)


copyparty/web/browser.js
```diff
@@ -9393,6 +9447,198 @@ ebi('files').onclick = ebi('docul').onclick = function (e) {
...
+		path: null,
+		id: null,
+		relpath: null,
+		no_dsel: false
+	};
+
+	function mktemp(is_dir) {
+		let row = mknod('tr', 'temp', 
+			'<td>-new-</td>' +
+			'<td columnspan="2"><input id="tempname" class="i" type="text" placeholder="' + (is_dir ? 'Folder' : "File") + ' Name"></td>'
```
Scotsguy: did you mean `colspan`?

9001:
nice!! struggling to even find nitpicks, this is dope :grin: 

only thing is the use of `let` rather than `var`, do you think a find/replace to use `var` instead would cause issues? ah, I guess `let fn = function() {};` would need to be scoped inside an iife or something... (sorry)

copyparty/web/browser.js
```diff
@@ -9393,6 +9447,198 @@ ebi('files').onclick = ebi('docul').onclick = function (e) {
...
+			if (!name)
+				return;
+			let data = new FormData();
+			data.set("act", is_dir ? "mkdir" : "new_md");
+			data.set("name", name);
+
+			let req = new XHR();
+			req.open("POST", get_evpath());
+			req.onload = req.onerror = function() {
+				if (req.status != 201)
```
Scotsguy: there's lots of errors that could reasonably be handled here, at least by passing on the error message. e.g. "that file already exists"
also, `new_md` responds with HTTP 200 on success (arguably incorrect), so this will show an error and not reload even when everything worked fine

copyparty/web/browser.js
```diff
@@ -640,6 +643,22 @@ if (1)
 		"ur_um": "Finished;\n{0} uploads OK,\n{1} uploads failed, sorry",
 		"ur_sm": "Finished;\n{0} files found on server,\n{1} files NOT found on server",
 
+		"rc_opn": "Open",
+		"rc_ply": "Play",
+		"rc_pla": "Play as audio",
+		"rc_txt": "Open in text editor",
```
Scotsguy: technically, this opens the text file viewer, not the editor. the viewer then has a button to open the md editor

copyparty/web/browser.js
```diff
@@ -9393,6 +9447,198 @@ ebi('files').onclick = ebi('docul').onclick = function (e) {
...
+				break;
+			case 'pst':
+				fn = function() {fileman.paste(); fileman.clip = []};
+				break;
+			case 'nfo':
+				fn = function() {mktemp(true)};
+				break;
+			case 'nfi':
+				fn = function() {mktemp()};
+				break;
```
Scotsguy: in firefox, neither new file nor new folder actually add the file/folder to the list until after a reload; they just disappear after pressing enter

copyparty/web/browser.js
```diff
@@ -9393,6 +9447,198 @@ ebi('files').onclick = ebi('docul').onclick = function (e) {
...
+
+		let input = ebi("tempname");
+		input.onblur = function() {
+			sendit(input.value);
+			row.remove();
+		};
+		input.onkeydown = function(e) {
+			if (e.key == "Enter")
+				sendit(input.value);
+			if (e.key == "Enter" || e.key == "Escape")
```
Scotsguy: pressing escape to cancel creating a new file also closes the top menu (up2k/settings/etc)

copyparty/web/browser.js
Scotsguy: in chromium, this happens:
image

Scotsguy (changes requested):

copyparty/web/browser.js
stackxp: Resolved by capturing the event

copyparty/web/browser.js
stackxp: Dunno what you mean, iT wORkS oN mY MAchInE!1

copyparty/web/browser.js
stackxp: Fixed!

copyparty/web/browser.js
stackxp: Added a few extra cases!

copyparty/web/browser.js
stackxp: Did that!

copyparty/web/browser.js
Scotsguy: > Dunno what you mean, iT wORkS oN mY MAchInE!1

do you mean the bug in firefox or the exception in chromium?

copyparty/web/browser.js
```diff
@@ -9393,6 +9447,202 @@ ebi('files').onclick = ebi('docul').onclick = function (e) {
...
+				return;
+			let data = new FormData();
+			data.set("act", is_dir ? "mkdir" : "new_md");
+			data.set("name", name);
+
+			let req = new XHR();
+			req.open("POST", get_evpath());
+			req.onload = req.onerror = function() {
+				if (req.status == 405 || req.status == 500)
+					return toast.err(3, "a " + (is_dir ? "folder" : "file") + " with that name alredy exists.");
```
Scotsguy: "already"

copyparty/web/browser.js
```diff
@@ -9393,6 +9447,202 @@ ebi('files').onclick = ebi('docul').onclick = function (e) {
...
+		path: null,
+		id: null,
+		relpath: null,
+		no_dsel: false
+	};
+
+	function mktemp(is_dir) {
+		let row = mknod('tr', 'temp', 
+			'<td>-new-</td>' +
+			'<td colspan="' + (QSA("#files thead th").length - 1) + '"><input id="tempname" class="i" type="text" placeholder="' + (is_dir ? 'Folder' : "File") + ' Name"></td>'
```
Scotsguy, line 9466: nitpick, but this is now so wide that the margin between textfield and containing box on the right side is smaller than the margin on the left

ex:
Image

Scotsguy (changes requested):

copyparty/web/browser.js
stackxp: > do you mean the bug in firefox or the exception in chromium?

On my computer, using Firefox. Apart from that, I think what might be causing the issue is that copyparty isn't fast enough to create the file/dir before the browser gets reloaded

Scotsguy: > On my computer, using Firefox. Apart from that, I think what might be causing the issue is that copyparty isn't fast enough to create the file/dir before the browser gets reloaded

for me it doesn't actually reload by itself, i have to do that manually
and i dont think performance is the issue necessarily, copyparty is writing to a plenty fast SSD in this case

stackxp: It now reloads after a delay of ~100ms which should be enogh, also `reload_browser()` doesn't actually reload anything lol.

9001: The response to the mkdir request isn't sent until the action has completed, so assuming RCM waits for the response before reloading then no manual delay should be necessary-- can you check in the browser network tab to ensure it does?

stackxp: I checked, this also works. So it _was_ just the `reload_browser()`-part that wasn't working.

copyparty/web/browser.js
```diff
@@ -9638,14 +9638,17 @@ var rcm = (function () {
 		menu.style.display = '';
 	}
 
+	var lrcTime = null;
 	ebi('wrap').oncontextmenu = function(e) {
 		hide(true);
-		console.log(e.detail)
-		if (thegrid.en || !r.enabled || (r.double && e.detail > 1))
+		if (thegrid.en || !r.enabled || (r.double && lrcTime && e.timeStamp - lrcTime < 500)) {
```
Scotsguy: this will open the native right click menu only when double-clicking quickly. I'm not sure that's what I would expect as a user
for example in youtube, right clicking anywhere [in the video player] while the custom rclick menu is open will open the native one, regardless of how long it's been since opening it initially

I think making it time-independent would be a good change for accessibility and how obvious the UI is to interact with

stackxp: Fixed by using the shift key to show the normal context menu (-1 translation key :partying_face:)

Scotsguy: I'm not sure that's necessarily better? what was the bug you fixed in [7a05108](https://github.com/9001/copyparty/pull/1135/commits/7a05108a5114e4f6852a261126cc4b6e97ae494e)?

stackxp: > what was the bug you fixed in 7a05108

The bug I fixed was that on Chromium, the double-click behavior didn't work because it doesn't set the `detail` property on `contextmenu` events, which shows single, double and triple clicks. (So actually Firefox is more out of spec)
That was just my own implementation of a double right-click which, to be fair, may not be the best practice.

> I'm not sure that's necessarily better?

I'm no UI designer, but I found the YouTube behavior quite annoying for copyparty (imo) so I thought that putting the hint in the context menu for the settings page was enough.
However, if you think that'd be better then it's ok too ¯\\\_(ツ)\_/¯

Scotsguy: i think we're talking past each other here: for the behaviour im imagining it doesn't matter what properties the contextmenu event has. the event handler goes as follows: if the rclick menu is currently hidden, show it. if it is currently visible, ignore the event (and show the native rclick menu). i admit i didnt inspect your implementation very closely at that point

i believe you're thinking of double right-clicking like double left-clicking, i.e. twice in quick succession, right?

as for shift clicking, probably just leave that in, and without a toggle
nb: firefox will never send a contextmenu event when shift-clicking (i.e. will always open the native rclick menu when shift-rclicking), so this would just mirror that behaviour to other browsers

stackxp: > i think we're talking past each other here: for the behaviour im imagining it doesn't matter what properties the contextmenu event has.

Okay, so with [b52aac1](https://github.com/9001/copyparty/pull/1135/commits/b52aac1a11d04922762cad4e3e8bc1a709f44c08) I removed the double-right-click functionality and replaced it with the shift-right-click thingy.

> i believe you're thinking of double right-clicking like double left-clicking, i.e. twice in quick succession, right?

Yes, that's what I meant. (that probably also shows it's not the most user-friendly option)

> as for shift clicking, probably just leave that in, and without a toggle

Makes sense, because it's already on Firefox.

So from what I've understood you want the youtube-like behavior and the shift-right-click as a polyfill for Chromium?

Scotsguy: > So from what I've understood you want the youtube-like behavior and the shift-right-click as a polyfill for Chromium?

not sure if polyfill is the right word for it, but basically yeah

(sorry, haven't gotten around to testing this yet)

PR #1151, opened 2025-12-29
u2c: add --ba argument to use basicauth instead of PW header

### LINKED ISSUE #1150: u2c: add an option to use basic-auth
Basically I am doing the same thing as #38 where I have a reverse proxy in front of copyparty (traefik) which performs auth using basicauth.

This works fine, however, u2c.py is using the PW header for sending the password:
https://github.com/9001/copyparty/blob/4642d32366e30c131d933c8bd0a519b69f4d3fff/bin/u2c.py#L197-L198

This means traefik is rejecting the request with a 401 :(

Could we have a u2c option to instead use basic auth?

I can PR this, but I'm not sure what to call the option (`--ba`?)

<hr>

p.s. should probably expand the 403 badauth check to also include 401 at https://github.com/9001/copyparty/blob/4642d32366e30c131d933c8bd0a519b69f4d3fff/bin/u2c.py#L860-L862

Le0Developer:
This allows using u2c.py with a copyparty server behind a reverseproxy which enforces authentication through basicauth

Also expands the BadAuth check to status 401.

Closes #1150

I hope this is compatible with python 2.6, it's been some time…

### DIFF
--- a/bin/u2c.py
+++ b/bin/u2c.py
@@ -195,7 +195,10 @@ class HCli(object):
 
         hdrs.update(self.base_hdrs)
         if self.ar.a:
-            hdrs["PW"] = self.ar.a
+            if self.ar.ba:
+                hdrs["Authorization"] = "Basic " + base64.b64encode(self.ar.a.encode("utf-8")).decode("utf-8")
+            else:
+                hdrs["PW"] = self.ar.a
         if ctype:
             hdrs["Content-Type"] = ctype
         if meth == "POST" and CLEN not in hdrs:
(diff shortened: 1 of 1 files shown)


9001:
Thanks! Made some small changes to make it possible to individually specify a separate password for basic-auth, in case someone has a setup like that. Did I mess anything up or does this still cover your usecase as well?

Le0Developer:
Thanks, this will work as well!

9001:
sweet, thx again :grin:

PR #1159, opened 2026-01-02
Fix right-click menu

stackxp:
Fixes a crash involving the right-click menu when the file type isn't the fourth column of the file type.
(Very specific I know)

(Also, this PR complies with the DCO; https://developercertificate.org/)


9001:
thx o/

PR #1160, opened 2026-01-02
Improve right-click menu (again)

stackxp:
- Added "copy link" option to menu
- Corrected "file viewer" to "text file viewer" and "text-editor" to "markdown editor"
- made the code a bit more compact


9001:
I think text-editor is a better description for it, since it can be used to edit arbitrary files as well, and not just markdown -- the preview pane doesn't really do much, but it still works :grin: 

aside from that, looks good

pyccl:
I also think text-editor is better.

PR #1164, opened 2026-01-03
dropdown grid view fix

Foox-dev:
Added the ability for the dropdown menu to work in the grid view. Before it seemed that it was only available in the list view, and had no functionality in grid view.


9001:
neat! thanks o/

PR #1167, opened 2026-01-04
Fix reverse proxy argument in nginx config

Raphdf201:


9001:
heh thanks :grin:

PR #1184, opened 2026-01-09
Fix and improve right-click menu (once again)

stackxp:
This fix improves some things and fixes some bugs with the right-click menu:
- Added an option to enable the youtube-like behavior (when you right-click and it's already open, use the browser's menu), defaults to off because it was kind of annoying \:)
- Added an option in the menu to rename files (thanks @feenix1)
- Changed the "text editor" option to "markdown viewer" (please tell me if there's a better option, like adding two editors or smth)
- Fixed some permission errors with the menu (it showed the delete option if you only had the move permission)
- Added support for filkeys and dirkeys (thanks @Emptiest)
- Fixed a visual bug where the menu would show a seperator at the top that seperates nothing if you only had the read permission


9001:
> Added support for filkeys and dirkeys

looks like i beat you to this one -- hope you don't run into too many issues when rebasing!

stackxp:
I hope that fixed it \:)

9001:
> when you right-click and it's already open, use the browser's menu), defaults to off

hehe thx for the motivation to add a global-option to set the default :grin: 

> Changed the "text editor" option to "markdown viewer" (please tell me if there's a better option, like adding two editors or smth)

right, this is probably yet another thing that's poorly documented -- a user who has write-delete in a volume is able to open any file in the editor by adding `?edit` to a URL; that editor was originally called the "markdown editor" and that's still its main purpose, but it can also do general-purpose editing of html and such, so perhaps "text-editor" is a better name now... ehh idk :p 

though if someone who doesn't have delete-access, the can only edit `.md` files, and trying to edit a non-md just gets a download

do you want to change anything before we merge this?

stackxp:
I changed the text from "markdown editor" to "markdown viewer" because adding `?v` to the URL of a markdown file opens not an editor, but a viewer from which you can select one of the two editors. I found that more fitting and you can still open one editor from the textfile viewer so I thought that might be the best option.

9001:
aight, thx again o/

stackxp:
np :)

copyparty/web/browser.js
```diff
@@ -9598,6 +9602,7 @@ var rcm = (function () {
 				}
 			}
 		}
+		console.log(selFile);
```
Scotsguy, line 9605: rogue `console.log()´?

PR #1193, opened 2026-01-11
nixos: use mkPackageOption instead of mkOption for the package option

toast003:
This gets rid of a warning when trying to build a system with `documentation.nixos.includeAllModules` enabled
```log
warning: Attempt to evaluate package pkgs.copyparty.out in option documentation; this is not supported and will eventually be an error. Use `mkPackageOption{,MD}` or `literalExpression` instead.
```

### DIFF
--- a/contrib/nixos/modules/copyparty.nix
+++ b/contrib/nixos/modules/copyparty.nix
@@ -69,11 +69,8 @@ in
   options.services.copyparty = {
     enable = mkEnableOption "web-based file manager";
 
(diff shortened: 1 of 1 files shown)


9001:
thx o/

toast003:
Actually no it doesn't, I just tested it on my pc which does not have `documentation.nixos.includeAllModules` enabled :)
Oh well, this is more correct so it doesn't matter

PR #1198, opened 2026-01-13
fix: make right-click menu work in search view

hackysphere:
This should fix the bug screen showing up when right-clicking in search results. The music playback and text editor context options should be functional, although markdown and text files won't have the markdown editor option because the leading column doesn't have a `-txt-` button.
(if there is something wrong with my code quality or implementation, please tell me, as this is my first contribution)

### DIFF
--- a/copyparty/web/browser.js
+++ b/copyparty/web/browser.js
@@ -9500,9 +9500,10 @@ var rcm = (function () {
 		elem: null,
 		type: null,
 		path: null,
+		dpath: null,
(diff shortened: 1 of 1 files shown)


hackysphere:
Added a change in ac2fbac that makes the option to open in the textfile viewer show for anything that is not a directory, pretty sure that is how the UI worked before as well.

9001:
nice! thanks :>

PR #1214, opened 2026-01-16
work out dsel bugs

icxes:
- also probably introduce new bugs
- also work out merge conflicts
- also cleanup things
- also tabs instead of spaces

this diff is sponsored by vscode, i have to test this a bit but thought i'd put it in so you won't go around changing it even more yet (it was already hard resolving the merge conflicts)

bugs found and fixed hopefully:
- couldn't select text
- didn't really work on mobile (kinda does now)
- better selection
- 

### DIFF
--- a/copyparty/web/browser.js
+++ b/copyparty/web/browser.js
@@ -9790,9 +9790,7 @@ function reload_browser() {
 		return { x: touch.clientX, y: touch.clientY };
 	}
 
-	function sel_toggle(el, usel) {
-		var m = usel ? true : 't';
-
+	function sel_toggle(el, m) {
 		clmod(el, 'sel', m);
 		var eref = el.getAttribute('ref');
 		if (eref) {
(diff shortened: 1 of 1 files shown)


9001:
> it was already hard resolving the merge conflicts

dw i'll be stuck on the office most of the day so that won't be a problem this time ( ´ w `)

icxes:
- holding down shift while releasing LMB will add to selection instead of toggling
- holding down alt while releasing LMB will remove from selection instead of toggling
- can select single item
- less dead code

can't think of anything to add for now so i guess it's good to go

9001:
the way file selection works in the listview is that if you start selecting from an item which is unselected, then every item you intersect with gets selected -- and vice versa, start from a selected item and everything you swipe past gets unselected -- using shift-cursorkeys I mean

would that make sense for gridview as well?

9001:
ah maybe not, because you can start selections from the blank area , and not necessarily starting from a file

icxes:
i would probably just keep it similar to how you would expect a file manager to work (which it admittedly maybe isn't right now, since the normal non-modifier action "should" be to only select what's covered by the selection and unselect anything else), but yeah without enough area to start the selection from it starts feeling a bit clunky

9001:
I'll say I'm a big fan of being able to easily add/subtract stuff from the selection, and (in line with everything else) not too worried about whether its intuitive or not, more important that it's convenient/powerful once you're used to it -- so TLDR I'll leave it to you w

PR #1220, opened 2026-01-18
reverse proxy .confs to trust docker daemon ip + cf real-ip

gkp1:
- working example for reverse proxy + docker AND cloudflare proxy + 'show real ip' config

1. .conf file tells copyparty to trust the docker daemon (docker host?) ip and rproxy -1 to grab the right ip

- this was really hard to figure out and debug as copyparty does not trust docker daemon ip so that's what is missing for docker + reverse proxy + cloudflare proxy setup

- specifically the 'trust docker ip' config with a .conf file is missing from docs here, and the proper .conf syntax, as the app console warnings assume you're running it directly in cli and only recommends the cli commands to use (rfproxy, xff-src, xff-hdr) and not the .conf syntax so I felt kinda lost on that until I figured it out, and the copyparty app just breaks if you put a command in the docker compose 'command' field, so this could certainly be useful to people setting this up


gkp1:
Edit: I noticed the nginx example is redundant cuz the example from the repo already has the working cloudflare config example -  Just deleted it. 

If you want, could merge only the .conf example to trust docker daemon ips, and to use -1 rfproxy config, 'real ip' won't be trusted by copyparty and https domain + cf won't work well in docker otherwise. People will definitely be stuck like me for a while or use a 'broken' docker setup with nginx if we don't add this example. 😅

9001:
Before I comment on the PR itself, please have some of my thoughts (and despair) on the topic.

Every time I set up a new webserver, one of *the* first things I do is to ensure that the correct IP-addresses are logged -- in every service, for every request. This is *essential* for debugging and diagnostics -- why would you troubleshoot something when you're not even able to track and correlate the requests which are encountering issues? So of course I made the assumption that everyone else did this as well, because surely nobody would want to run their stack blindly...? And so, for the longest time, copyparty did exactly what you're suggesting in this PR: it assumed some sane defaults, and that was that.

Except, it turns out that many people never configured their reverseproxies correctly, leading to some massive headaches. Let's look at some examples that repeatedy appear in the github issues:

* barrage of malicious requests from an IP? No problem; let's just ban the IP-address. Except, wait, that was the reverseproxy's LAN IP! Now *everyone* is banned!

* someone recently uploaded a file by accident, and now they want to delete it with the unpost feature. No problem; we know what IP they uploaded from, so let's just give them the shiny delete-button... Except, wait, that was the reverseproxy's LAN IP, and now *everyone* can delete *any* recent upload!
  * this one is "only" a problem on servers with shared accounts, which appears to be a fairly popular way to run copyparty

* the server has IdP-based authentication, and the reverseproxy has added the client's username in the correct header, so all is good -- let them in. Except, wait, that request never even went through a reverseproxy at all, and the client is spoofing somebody else!
  * okay, this one isn't IP-related, but still a scary example of misconfigured reverseproxies, (EDIT:) and is also not a real example, not something I know someone has actually bumbled into -- but still a scary thought 

so it is unfortunately true that it is *absolutely not* possible to set default values for these properties in the config. Each environment that copyparty runs in is, sadly, vastly different. If only people would sanitize their x-forwarded-for headers (something I naively expected) then we wouldn't be having this issue at all; alas, those are not the cards we are dealt.

So while it is already obvious that we cannot provide any default values for these, let me also give some examples why the particular defaults you selected are problematic and could not be *the* defaults:

* `xff-src: 10.0.0.0/8` will only work for a subset of deployments, where the reverseproxy is another container in the same container network --
  * if the reverseproxy is running on the host, it does not work
  * if the reverseproxy is running on another machine in the LAN, it does not work
  * if there is no reverseproxy, it does not work
  * if the container is running with host networking, it does not work
* `rproxy: -1` assumes the `x-forwarded-for` header is safe to use --
  * assumes the reverseproxy has even bothered verifying that the values it provided in the header are trustworthy; empirically a bad assumption
  * assumes the reverseproxy even bothered doing anything at all about the header, empirically a bad assumption
  * disables the configuration advice message in the copyparty log

I think the correct approach would be to add some comments instead, pointing out that the copyparty log will have further advice on how to configure it:

527340125-35a809e9-e117-478b-83bd-95eed2682a2e

There should also be a warning against "just picking something that seems to work"; specifically, `rproxy` in particular must NEVER be configured by default in ANY example config, because it disables the above message, masking the problem instead of solving it, leading to further despair on all sides of the fence (you, me, hosters, visitors).

gkp1:
@9001 first off thanks for the response, I pushed this .conf addition as a 'default' example  because I assumed this example docker compose and conf file would be giving users a 'working example' of a docker compose setup, but I realize now that I assumed that the user would have nginx setup on the host like I did, and not inside the docker compose, or not use another reverse proxy, etc. Maybe this could be added as a comment or a suggestion in the .conf file, so the user could at least have another config example/direction to debug if they are getting the 'copyparty does not know which ip is the real user ip' issues with docker and they have no idea that they need to trust the docker daemon ips.

PR #1223, opened 2026-01-18
Random sorting

evergreenbear:
Hi,
I want to preface this by saying that this is both my first PR to any project and my first time doing anything with Javascript. However it is very much working (for me, at least) on both Firefox and Chromium.
image
This adds a new sorting option, random, that shuffles the contents of the directory in a unique order each time it is selected.

I imagine there are likely changes that still need to be made for this to be production-ready which I had not managed to catch, but I figured I got somewhere and this may still be useful.
I modified `make-sfx.sh` because cp would error on that command without  `-r` on the system I was using to build (AlmaLinux amd64), and the script would not finish.

### DIFF
--- a/copyparty/web/browser.js
+++ b/copyparty/web/browser.js
@@ -487,6 +487,7 @@ if (1)
 		"gt_sz": "size",
 		"gt_ts": "date",
 		"gt_ext": "type",
+		"gt_rand": "random",
 		"gt_c1": "truncate filenames more (show less)",
 		"gt_c2": "truncate filenames less (show more)",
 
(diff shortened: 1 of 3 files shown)


9001:
Hey :>

I'm curious to hear the reason you want to add this; what's the usecase?

evergreenbear:
Truthfully, my use case is to randomly sort albums of old photos to look through for fun, though I imagine this would be more useful if used in conjunction with a slideshow feature for e.g. turning an old device into a digital photo frame

overcastbulb:
This looks like a useful addition! Random sorting is a common ask for media-heavy use cases like photo browsing and digital photo frames. The implementation is small and focused.
A few things worth checking before merge:

Is the shuffle using Math.random() directly? If so, consider whether Fisher-Yates shuffle would be more appropriate for uniform randomness
Does the random order reset/reshuffle each time the option is selected, or does it persist?

Either way, great first PR @evergreenbear — would love to see this merged! 🎉

Kistras:
I would love to see this feature in copyparty and I really wanted it in my instance. Unfortunately it doesn't feel ready at the moment

I suggest detaching it from regular sorting so that you can't select "random" as primary and "name"/other as secondary sorting options. Sorting direction shouldn't matter for it
image

Table header in list view is broken: it's displayed in an incorrect place. Hiding it hides "type" column instead. I believe it shouldn't appear as a header at all (more points to it being a separate button)
image

Random sorting doesn't apply upon reloading the page or loading more items if a directory contains a lot of items. If it *were* to apply, I suggest storing a seed somewhere and using that so that sorting is reproduceable (with something like sfc32: https://stackoverflow.com/a/47593316)

PR #1225, opened 2026-01-19
Document Cloudflare proxy upload speed optimization

gkp1:
Added instructions for fixing upload speed and **big uploads not working in general** when using Cloudflare proxy with nginx.

### DIFF
--- a/contrib/nginx/copyparty.conf
+++ b/contrib/nginx/copyparty.conf
@@ -32,7 +32,18 @@
 #   (curl -s https://www.cloudflare.com/ips-v{4,6} | sed 's/^/allow /; s/$/;/'; echo; echo "deny all;") > /etc/nginx/cloudflare-only.conf
 #
 # and then enable it below by uncommenting the cloudflare-only.conf line
+# 
 #
+# ** Fix upload speed for Cloudflare proxy: from <1MBps to >40MBps **
+# If using cloudflare proxy, replace: 	
+# proxy_set_header   Connection        "Keep-Alive";
+# with:
+# proxy_set_header   Connection        "";
+# and just make sure you also have: proxy_http_version 1.1;
+#
+# Reason: Clears client "Connection" headers so Nginx can manage its own
+# persistent pool via the "upstream" keepalive. Prevents Cloudflare/Client
+# close-signals from killing the backend pipe, drastically boosting speed.
 # ======================================================================
 
 


gkp1:
I also commented out the `#client_max_body_size 1024M;` line at the end. I kept `client_max_body_size 0;` in the middle. not sure if this was duplicated accidentally? Nginx complains about it.

Maybe the above 1024M limit should be commented out by default too? Not sure though.

More info: this was really bizarre fix. I could not understand why I had unusable upload speed with cloudflare proxy. Turns out that was literally the only thing I needed to change. NOW my upload speed with https domain + cloudflare proxy  is the same as using the "naked" ip! (which is surprising?) I have 40MBps upload speed here on both tests. This is how slow it was before I did this change:

image

gkp1:
@9001 if you want I can add a quick video switching this config and testing, if would make it easier for you to confirm it

9001:
huh? this config change and the comments within does not make any sense;

> Clears client "Connection" headers so Nginx can manage its own persistent pool via the "upstream" keepalive. 

the http spec clearly states that the default value of the `Connection` header is `keep-alive` when using `http/1.1`, so this does not actually change anything at all. This is also trivial to verify with the copyparty global-option `ihead: connection`.

further,

> Prevents Cloudflare/Client close-signals from killing the backend pipe, drastically boosting speed.

this nginx config-option only affects communication between nginx and copyparty, and is entirely unrelated to communication towards the client/cloudflare side. Why do you believe it "Prevents Cloudflare/Client close-signals"?

for good measure, I did try your suggested config-change on one of my own server, which has cloudflare -> nginx -> copyparty running directly on the host OS (alpine) -- just as expected, it had zero impact on upload speed. And yes, I did confirm that the config-change was applied, since copyparty was no longer receiving the "Connection" header.

can you explain how your reasoning and those explanations came to be?

gkp1:
> > Clears client "Connection" headers so Nginx can manage its own persistent pool via the "upstream" keepalive.
> 
> the http spec clearly states that the default value of the `Connection` header is `keep-alive` when using `http/1.1`, so this does not actually change anything at all. This is also trivial to verify with the copyparty global-option `ihead: connection`.
> 

Hi 9001, sorry for the misunderstanding here. I certainly misdiagnosed the issue. Confirmed what you said and it was neither max body size or the keepalive value, as you said. Also I do not have deep knowledge on the http spec, I was just switching the valid configs for Connection.

What misled me was prob. the order which I setup my nginx. I had issues setting up the nginx file by starting with your nginx example, so I started with my default nginx file I use for all my compose apps (mostly the same as yours without all the performance specific configs). 

After it was working, I merged all your extra nginx performance configs into my base nginx file, replacing anything conflicting. Added cpp_tcp etc.

What "fixed my cloudflare performance issue" was definitely applying all your examples to my nginx file cpp_tcp, proxy_request_buffering, buffer size, redirect etc, honestly no idea why they did not work immediately, even after applying the correct performance configs & restarting nginx several times and testing - then I randomly tried the "stupid fix" which changed nothing like you explained and since it started working I had the placebo feeling that it fixed uploads. lol. Thanks for the explanation though.

PR #1229, opened 2026-01-21
Esperanto translation update

slashdevslashurandom:
Important changes:

The "shift" key is called "MAJ" (short for "majuskla" for "uppercase") in the translation, so the new lines have been updated to match.

"beligi json" is technically a valid string (proper names from other languages don't have to be grammatically modified), but the "-on" ending *could* be confused for a grammatical ending, with the whole string being interpreted as "beautify javascript" instead, so instead "JSON" is capitalized and another "on" ending is added.

A few commas are added for readability.


9001:
awesome, thanks for fixing it up :>

PR #1238, opened 2026-01-24
update web dependency - prism

grawlinson:
* Update prism from 1.29.0 to 1.30.0.
* Separate the plugins into their own array (similar to langs).
* Add some more langs to highlight.


9001:
Thanks!

Though I don't think it's a good idea to enable all the languages; it brings the size of the prism build up from 48 to 208 KiB, which would increase the total size of the JavaScript that gets loaded when loading the web-UI from 168 to 329 KiB. 

but mentioning all the languages in the list is a good idea for context, so how about making the languages you added comments instead?

```bash
langs=(
    markup
    css
    clike
    javascript
    # abap
    # abnf
    # ...
```

grawlinson:
Sounds good to me!

PR #1240, opened 2026-01-24
fix: update `utime_c` to accept float timestamps

### LINKED ISSUE #1239: WebDAV `x-oc-mtime` header fails on decimal timestamps
### Describe the bug

Some WebDAV clients (notably Tampermonkey sync) send `x-oc-mtime` as a decimal Unix timestamp (e.g., `1632900592.112`). Copyparty currently parses this value as an `int`, which raises a `ValueError` and rejects the upload.

```
invalid literal for int() with base 10: '1632900592.112'

URL: userscripts1/Tampermonkey/sync/ab2e543d-a275-40e7-8d05-b4f4c8fe398f.meta.json
```

### To Reproduce

```
curl 'https://<redacted>.com/userscripts/Tampermonkey/sync/ab2e543d-a275-40e7-8d05-b4f4c8fe398f.meta.json' \
  -X 'PUT' \
  -H 'authorization: Basic <redacted>' \
  -H 'content-type: application/octet-stream' \
  -H 'x-oc-mtime: 1597751672.426' \
  --data-raw '{"uuid":"ab2e543d-a275-40e7-8d05-b4f4c8fe398f","name":"My Script","options":{},"lastModified":1597751672426}'
```

### Expected behavior

Upload succeeds, and the file mtime is set to `1597751672.426` (sub-second precision preserved if supported by the filesystem/OS).

### Server details (if you are using docker/podman)
* **how you're running copyparty:** docker
* **docker image:** ghcr.io/9001/copyparty-ac, amd64, latest

### Server details (if you're NOT using docker/podman)
* run copyparty with `--version` and grab the last 3 lines (they start with `copyparty`, `CPython`, `sqlite`) and paste them below this line:
```py
copyparty v1.20.4 "sftp is fine too" (2026-01-23)
  CPython v3.12.12 on Linux64  [GCC 15.2.0]
   sqlite 3.51.2*1 | jinja 3.1.6 | pyftpd 1.5.10 | tftp 0.4.0 | miko 4.0.0
```

### Server log

```py
00:05:19.938 192.168.1.180 56866   PUT  /userscripts/Tampermonkey/sync/ab2e543d-a275-40e7-8d05-b4f4c8fe398f.meta.json @userscripts
00:05:19.943 192.168.1.180 56866   http500: httpcli.py:827 <run>: return self.handle_put() and self.keepalive
httpcli.py:2164 <handle_put>: return self.handle_stash(True)
httpcli.py:2671 <handle_stash>: post_sz, halg, sha_hex, sha_b64, remains, path, url = self.dump_to_file(is_put)
httpcli.py:2556 <dump_to_file>: bos.utime_c(self.log, path, int(cli_mt), False)
[ValueError] invalid literal for int() with base 10: '1667526220.297', '/userscripts/Tampermonkey/sync/ab2e543d-a275-40e7-8d05-b4f4c8fe398f.meta.json'
```

### Additional context

I've created a PR to address the issue #1240

jcwillox:
Some WebDAV clients (e.g., Tampermonkey sync) send `x-oc-mtime` as a decimal string. Copyparty previously used `int(...)`, which raises `ValueError` for decimal inputs. Python’s `os.utime` accepts float timestamps, so it’s safe to pass floats and preserve sub-second precision where supported.

I updated `bos.utime_c` to accept float timestamps, and added a test for when `x-oc-mtime` is in milliseconds, and parse the header as a float now.

However if you prefer simpler approach, you could also just change `int(cli_mt)` in `httpcli.py` to `int(float(cli_mt))` or `int(cli_mt.split('.')[0])` to truncate it. We could also try-catch to parse int first and fall back to float only on failure.

Let me know if you want any changes, or a different approach

Fixes #1239

### DIFF
--- a/copyparty/httpcli.py
+++ b/copyparty/httpcli.py
@@ -2618,7 +2618,7 @@ class HttpCli(object):
         at = mt = time.time() - lifetime
         cli_mt = self.headers.get("x-oc-mtime")
         if cli_mt:
-            bos.utime_c(self.log, path, int(cli_mt), False)
+            bos.utime_c(self.log, path, float(cli_mt), False)
 
         if nameless and "magic" in vfs.flags:
             try:
--- a/copyparty/bos/bos.py
+++ b/copyparty/bos/bos.py
@@ -106,14 +106,14 @@ def utime(
 def utime_c(
     log: Union["NamedLogger", Any],
     p: str,
-    ts: int,
+    ts: float,
     follow_symlinks: bool = True,
     throw: bool = False,
-) -> Optional[int]:
+) -> Optional[float]:
     clamp = 0
     ov = ts
     bp = fsenc(p)
-    now = int(time.time())
+    now = time.time()
     while True:
         try:
             if SYMTIME:
--- a/tests/test_webdav.py
+++ b/tests/test_webdav.py
@@ -63,6 +63,20 @@ Accept-Encoding: gzip
 fgsfds"""
 
 
+RCLONE_PUT_FLOAT = """PUT /%s HTTP/1.1
+Host: 127.0.0.1:3923
+User-Agent: rclone/v1.67.0
+Content-Length: 6
+Authorization: Basic azp1
+Content-Type: application/octet-stream
+Oc-Checksum: SHA1:f5e3dc3fb27af53cd0005a1184e2df06481199e8
+Referer: http://127.0.0.1:3923/
+X-Oc-Mtime: 1689453578.123
+Accept-Encoding: gzip
+
+fgsfds"""
+
+
 # tcpdump of `rclone delete dav:/a/d1/`  (it does propfind recursively and then this on each file)
 # (note: `rclone rmdirs dav:/a/d1/` does the same thing but just each folder after asserting they're empty)
 RCLONE_DELETE = """DELETE /%s HTTP/1.1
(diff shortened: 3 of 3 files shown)


copyparty/httpcli.py
```diff
@@ -2618,7 +2618,7 @@ def dump_to_file(self, is_put: bool) -> tuple[int, str, str, str, int, str, str]
         at = mt = time.time() - lifetime
         cli_mt = self.headers.get("x-oc-mtime")
         if cli_mt:
-            bos.utime_c(self.log, path, int(cli_mt), False)
+            bos.utime_c(self.log, path, float(cli_mt), False)
```
jcwillox, line 2621: could be a bit extra here and add a try-catch to ignore mtime when invalid, e.g.

```py
    def _parse_x_oc_mtime(self, raw: str) -> Optional[float]:
        try:
            return float(raw)
        except ValueError:
            self.log("invalid x-oc-mtime header %r" % (raw,), 6)
            return None
```

but idk if that's necessary

9001:
looks good as-is, thanks! probably healthy that it explodes loudly if an unsupported x-oc-mtime is provided, otherwise we might never notice :+1:

jcwillox:
sweet thanks 👍

PR #1244, opened 2026-01-24
sample rpxy configuration and test script

akostadinov:
Decided to contribute [rust-rpxy](https://github.com/junkurihara/rust-rpxy) config which is a very fast rust based proxy.

Will be very insightful to see how it fairs with the rest of the bunch. Author claims it being faster than nginx and caddy. 

The configuration file should be self-explanatory but there is a small readme and a test script to fill in details about how to run rpxy.

Regards.


9001:
Thanks -- looks interesting!

Just waiting for another issue to clear up first, since that could affect performance behind reverseproxies -- will need to grab new numbers after that anyways, so I'll wait until then.

PR #1252, opened 2026-01-26
Nix: package web dependencies attempt 2

toast003:
Better (hopefully) version of https://github.com/9001/copyparty/pull/897

Nix derivations run inside of a sandbox (think of docker, but more limited). Since we build web deps inside of a container anyways, this just extracts the commands from the Dockerfile so that we can run them inside the derivation for copyparty.

I didn't think this was feasible since you can't access the network from a normal derivation, until a while ago when I thought of using a [fixed output derivation ](https://bmcgee.ie/posts/2023/02/nix-what-are-fixed-output-derivations-and-why-use-them/), which can access the network.

This would require a lot less maintenance than the previous pr, however it could be problematic since Nix enforces that fixed output derivations are reproducible.


grawlinson:
Package maintainer from Arch Linux here, I was doing this from a different direction. :)

A few notes:

* zopfli/fonttools are available in alpine's repos (`py3-zopfli` and `py3-fonttools`).
* not building these from source means that cmake/others are no longer needed in the Dockerfile.
* `busy.mp3` is not reproducible at all (different output every time), which means that it's another candidate for vendoring.

I was hoping to decouple the downloading of sources from building, e.g. having a multi-layer Dockerfile like this:

1. install all deps from alpine repos.
2. download all archives & setup the files/dirs that the build scripts expect them to be in.
3. run `build.sh` similar to the one you currently have in this PR, with the ability to skip prebuilt/vendored deps.

9001:
sorry for the merge conflict; feel free to forcepush or rebase however you prefer :pray: 

EDIT: just to mention I'll be mostly off-grid until the weekend, got a webbrowser but no ability to build anything o/

toast003:
Dw @9001  , you're good!

> - zopfli/fonttools are available in alpine's repos (py3-zopfli and py3-fonttools).
> - not building these from source means that cmake/others are no longer needed in the Dockerfile.

What's your opinion on just using zopfli and/or fonttools from the repos? I can't really think of a reason to not use the repos, but since the Dockerfile gets specific versions I figured I might as well ask before doing anything

9001:
> What's your opinion on just using zopfli and/or fonttools from the repos? 

good idea, and probably overdue -- they weren't packaged by alpine back then is all 👍

grawlinson:
Definitely a lot better now. :)

9001:
LGTM! not much to say, happy to merge unless you had more changes in mind :+1:

toast003:
Hi, I have some unpushed changes pending that are stuck on my main PC (ram failed, and warranty process is slow), so I have been focusing on other things for now.

toast003:
That said I would be ok to merge this rn without commit 43d48b23fe407d894ee5095942ec59b95e26feb9 if splitting the dockerfile is useful for other packages, like @grawlinson said.

9001:
oof, dogspeed 🙏

and no rush of course o/

mid-kid:
gentoo packager here, I've [attempted](https://gist.github.com/mid-kid/cc7c0c2e1c188c8b135663d547e3dd35) building some of the webdeps outside of docker before, but concluded that the dependency tree of most of them was too deep. Still, this splitting effort was going to be how I wanted to tackle it if I'd kept at it, so I'm happy to see this.

maxprehl:
@toast003 any updates on this?

PR #1257, opened 2026-01-28
Fix `nth` and `nih`

stackxp:
This removes `nth` (which didn't actually do anything) and makes `nih` do what it should have done.
I tried to implement it as I think it was intended, please tell me if there are some other reasons that it was like this or if there are other changes to make.


9001:
thanks -- yeah `nth` was never implemented but `nih` works "as intended" -- it's mostly a cosmetic option to avoid taking up space in the UI, and `bname` is the intended way to also remove it from the title, so fixed the helptext of that too

PR #1259, opened 2026-01-29
Add checkconfig function to OpenRC script

lotsospaghetti:
I found it slightly cumbersome to write a config when the only feedback OpenRC gives is `crashed`, so I added a small QOL function to run copyparty with the `--exit cfg` flag directly from the service.


9001:
nice, thanks :>

PR #1265, opened 2026-02-01
add skip silence option

icxes:
adds skip silence toggle for audio tracks and some config options in 🎺

### DIFF
--- a/copyparty/web/browser.js
+++ b/copyparty/web/browser.js
@@ -327,7 +327,7 @@ if (1)
 		"mt_xowa": "there are bugs in iOS preventing background playback using this format; please use caf or mp3 instead",
 		"mt_tint": "background level (0-100) on the seekbar$Nto make buffering less distracting",
 		"mt_eq": "enables the equalizer and gain control;$N$Nboost &lt;code&gt;0&lt;/code&gt; = standard 100% volume (unmodified)$N$Nwidth &lt;code&gt;1 &nbsp;&lt;/code&gt; = standard stereo (unmodified)$Nwidth &lt;code&gt;0.5&lt;/code&gt; = 50% left-right crossfeed$Nwidth &lt;code&gt;0 &nbsp;&lt;/code&gt; = mono$N$Nboost &lt;code&gt;-0.8&lt;/code&gt; &amp; width &lt;code&gt;10&lt;/code&gt; = vocal removal :^)$N$Nenabling the equalizer makes gapless albums fully gapless, so leave it on with all the values at zero (except width = 1) if you care about that",
-		"mt_drc": "enables the dynamic range compressor (volume flattener / brickwaller); will also enable EQ to balance the spaghetti, so set all EQ fields except for 'width' to 0 if you don't want it$N$Nlowers the volume of audio above THRESHOLD dB; for every RATIO dB past THRESHOLD there is 1 dB of output, so default values of thresh -24 and ratio 12 means it should never get louder than -22 dB and it is safe to increase the equalizer boost to 0.8, or even 1.8 with ATK 0 and a huge RLS like 90 (only works in firefox; RLS is max 1 in other browsers)$N$N(see wikipedia, they explain it much better)",
+		"mt_drc": "enables the dynamic range compressor (volume flattener / brickwaller); will also enable EQ to balance the spaghetti, so set all EQ fields except for 'width' to 0 if you don't want it$N$Nlowers the volume of audio above THRESHOLD dB; for every RATIO dB past THRESHOLD there is 1 dB of output, so default values of 'tresh' -24 and 'ratio' 12 means it should never get louder than -22 dB and it is safe to increase the equalizer boost to 0.8, or even 1.8 with ATK 0 and a huge RLS like 90 (only works in firefox; RLS is max 1 in other browsers)$N$N(see wikipedia, they explain it much better)",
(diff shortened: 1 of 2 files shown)


copyparty/web/browser.js
```diff
@@ -326,7 +327,13 @@ if (1)
...
+		"mt_drc": "enables the dynamic range compressor (volume flattener / brickwaller); will also enable EQ to balance the spaghetti, so set all EQ fields except for 'width' to 0 if you don't want it$N$Nlowers the volume of audio above THRESHOLD dB; for every RATIO dB past THRESHOLD there is 1 dB of output, so default values of thresh -24 and ratio 12 means it should never get louder than -22 dB and it is safe to increase the equalizer boost to 0.8, or even 1.8 with ATK 0 and a huge RLS like 90 (only works in firefox; RLS is max 1 in other browsers)$N$N(see wikipedia, they explain it much better)",
```
```suggestion
		"mt_drc": "enables the dynamic range compressor (volume flattener / brickwaller); will also enable EQ to balance the spaghetti, so set all EQ fields except for 'width' to 0 if you don't want it$N$Nlowers the volume of audio above THRESHOLD dB; for every RATIO dB past THRESHOLD there is 1 dB of output, so default values of 'tresh' -24 and 'ratio' 12 means it should never get louder than -22 dB and it is safe to increase the equalizer boost to 0.8, or even 1.8 with ATK 0 and a huge RLS like 90 (only works in firefox; RLS is max 1 in other browsers)$N$N(see wikipedia, they explain it much better)",
```

copyparty/web/browser.js
```diff
@@ -2690,11 +2700,15 @@ var afilt = (function () {
 	var r = {
 		"eqen": false,
 		"drcen": false,
+		"ssen": false,
 		"bands": [31.25, 62.5, 125, 250, 500, 1000, 2000, 4000, 8000, 16000],
 		"gains": [4, 3, 2, 1, 0, 0, 1, 2, 3, 4],
 		"drcv": [-24, 30, 12, 0.01, 0.25],
-		"drch": ['tresh', 'knee', 'ratio', 'atk', 'rls'],
+		"drch": ['thresh', 'knee', 'ratio', 'atk', 'rls'],
```
```suggestion
		"drch": ['tresh', 'knee', 'ratio', 'atk', 'rls'],
```

copyparty/web/browser.js
9001: yeah I realize it's a typo, but it keeps the label in the UI short enough that it doesn't become wider than the inputbox and cause a reflow :p

copyparty/web/browser.js
```diff
@@ -2583,6 +2591,8 @@ var mpui = (function () {
 			// occasionally draw buffered regions
 			if (nth % 5 == 0)
 				pbar.drawbuf();
+
+			if (!IE && jread('au_ss', false)) skipSilence();
```
```suggestion
			if (!IE && afilt.ssen) skipSilence();
```

copyparty/web/browser.js
```diff
@@ -2769,6 +2783,7 @@ var afilt = (function () {
 			r.gains = gains;
 
 		r.drcv = jread('au_drcv', r.drcv);
+		r.sscv = jread('au_sscv', r.sscv)
```
```suggestion
		r.sscv = jread('au_sscv', r.sscv);
```

copyparty/web/browser.js
```diff
@@ -9929,6 +9984,94 @@ function reload_browser() {
 	dsel_init();
 })();
 
+function skipSilence() {
+	var ae = mp.au;
+	var ssconf = jread('au_sscv', [1, 5, 5, 5.0, 0.2]);
```
```suggestion
	var ssconf = afilt.sscv;
```

copyparty/web/browser.js
```diff
@@ -9929,6 +9984,94 @@ function reload_browser() {
...
+				ae.playbackRate += (tspeed - ae.playbackRate) * config.rspeed;
+			}
+			if (Math.abs(ae._gnod.gain.value - tvol) > 0.01) {
+				ae._gnod.gain.value += (tvol - ae._gnod.gain.value) * config.rspeed;
+			}
+		} else {
+			ae.playbackRate = 1.0;
+			ae._gnod.gain.value = 1.0;
+		}
+		setTimeout(detectSilence, config.loopInterval);
```
9001: iirc this eventually causes a stack overflow (or at least poor perofrmance); better to setInterval and then clearInterval at the return on line 10'031?

9001 (changes requested):
good stuff 👌

copyparty/web/browser.js
9001: though not too worried about that in the UI elements you added, worst case can compensate by growing the inputboxes a bit -- also feel free to add the finnish translations if you'd like

copyparty/web/browser.js
icxes: took a slightly different approach, think it'll work?

copyparty/web/browser.js
icxes: i'll check the mtl and fix if needed at some point

copyparty/web/browser.js
9001: yup! nice

9001:
sweet, thx -- I'll probably get a chance to play with this at the office tomorrow o/
* beta SFX and whl: https://copyparty.eu/beta/
* beta docker image: `docker/ac:beta` on [dockerhub](https://hub.docker.com/layers/copyparty/ac/beta/images)

also live on https://a.ocv.me/pub/demo/music/

PR #1266, opened 2026-02-02
web, th: rudimentary jxl support

### LINKED ISSUE #724: Support for JPEG XL (.jxl) image previews and thumbnails
**Is your feature request related to a problem? Please describe.**

Currently, Copyparty does not recognize or generate thumbnails for JPEG XL (`.jxl`) image files. When I upload `.jxl` files, they are treated as generic files rather than images. This means:

* No image thumbnails are generated in the file listing, making it difficult to browse visually.
* The images cannot be viewed directly in the browser by clicking on them; they must be downloaded first.
* This breaks the user experience for anyone using this modern and efficient image format, forcing them to either convert files to older formats or lose the convenient preview functionality that Copyparty provides for formats like JPEG, PNG, and GIF.

I would like Copyparty to have native support for the JPEG XL (.jxl) image format
(or optional support like many other features copyparty currently has)

This would involve:
1.  Recognizing `.jxl` files as a supported image type.
2.  Generating thumbnails for `.jxl` files to be displayed in the folder view.
3.  Enabling inline viewing of `.jxl` images in the browser when a user clicks on the file, similar to how JPEGs and PNGs are currently handled.

**Describe any alternatives you've considered**

The main alternative I'm aware of at the moment is to manually convert all ".jxl" files to a supported format like JPEG or PNG before uploading them to a Copyparty instance.

intelfx:
Fixes #724.

Not pretending this is mergeable as-is, I'm just putting it out here to be told where exactly I'm an idiot or something.

Ideally, I want to plumb progressive decoding through libvips for thumbnailing, to generate thumbnails with zero CPU load. If that's not crazy enough, we can even do thumbnailing via image truncation and avoid writing them to disk too.


copyparty/web/baguettebox.js
```diff
@@ -36,10 +36,10 @@ window.baguetteBox = (function () {
         touchFlag = false,  // busy
         scrollCSS = ['', ''],
         scrollTimer = 0,
-        re_i = /^[^?]+\.(a?png|avif|bmp|gif|heif|jpe?g|jfif|svg|webp)(\?|$)/i,
+        re_i = /^[^?]+\.(a?png|avif|bmp|gif|heif|jpe?g|jfif|svg|webp|jxl)(\?|$)/i,
```
9001:
```suggestion
        re_i = /^[^?]+\.(a?png|avif|bmp|gif|heif|jfif|jpe?g|jxl|svg|tiff?|webp)(\?|$)/i,
```

bin/hooks/reloc-by-ext.py
```diff
@@ -41,7 +41,7 @@
 """
 
 
-PICS = "avif bmp gif heic heif jpeg jpg jxl png psd qoi tga tif tiff webp"
+PICS = "avif bmp gif heic heif jpeg jpg jxl png psd qoi tga tif tiff webp jxl"
```
9001:
```suggestion
PICS = "avif bmp gif heic heif jpeg jpg jxl png psd qoi tga tif tiff webp"
```

9001: was already there 😁

copyparty/web/baguettebox.js
```diff
@@ -36,10 +36,10 @@ window.baguetteBox = (function () {
...
         re_v = /^[^?]+\.(webm|mkv|mp4|m4v|mov)(\?|$)/i,
         re_cbz = /^[^?]+\.(cbz)(\?|$)/i,
-        cbz_pics = ["png", "jpg", "jpeg", "gif", "bmp", "tga", "tif", "tiff", "webp", "avif"],
+        cbz_pics = ["png", "jpg", "jpeg", "gif", "bmp", "tga", "tif", "tiff", "webp", "avif", "jxl"],
```
9001:
```suggestion
        cbz_pics = ["png", "jpg", "jpeg", "gif", "bmp", "tga", "tif", "tiff", "webp", "avif"],
```

9001: I'm realizing cbz_pics is superfluous, can just use `re_i` the only place it is used (`fillCbzGallery`) -- I'll fix that 👍

copyparty/web/browser.js
```diff
@@ -1147,22 +1147,26 @@ function read_sbw() {
 onresize100.add(read_sbw, true);
 
 
-var have_webp = sread('have_webp');
-(function () {
-	if (have_webp !== null)
+function check_image_support(format, uri) {
+	var cached = sread('have_' + format);
```
```suggestion
	var cached = window['have_' + format];
```

copyparty/th_cli.py
```diff
@@ -89,10 +90,10 @@ def get(self, dbv: VFS, rem: str, mtime: float, fmt: str) -> Optional[str]:
 
         preferred = self.args.th_dec[0] if self.args.th_dec else ""
 
-        if rem.startswith(".hist/th/") and rem.split(".")[-1] in ["webp", "jpg", "png"]:
+        if rem.startswith(".hist/th/") and rem.split(".")[-1] in ["webp", "jpg", "png", "jxl"]:
```
9001:
```suggestion
        if rem.startswith(".hist/th/") and rem.split(".")[-1] in IMG_EXTS:
```

copyparty/th_cli.py
9001, line 24:
```suggestion


IMG_EXTS = set(["webp", "jpg", "png", "jxl"])
```

copyparty/th_cli.py
9001: black is gonna bite us here and reformat it (90chars or longer), and making it a set is better anyways

9001 (changes requested):
nice! on the contrary, I think this looks really good so far, just a few small nitpicks really :>

as for thumbnailing, while I think streaming truncated copies of the original images sounds fun, I suspect it would turn out messy and not as performant, so I'm leaning towards at least writing the truncated copies to the thumbnails-cache like the other formats.

copyparty/web/browser.js
intelfx: I changed this in a slightly different way.

intelfx:
All done. Unrelated changes were factored out in separate commits.

I also added a hack-ish thing to perhaps control libvips memory usage, because I noticed that copyparty eats a honestly obscene amount of RAM once you ask it to create a bunch of JXL thumbnails (open a grid with >9000 files in it, scroll to the end, enjoy 30+ GiB VIRT and 10+ GiB RSS on the copyparty process).

As far as I could observe, this is not a new behavior and it was happening all along (i.e. with webp thumbnails on the same grid I end up with 3-4 GiB RSS which is still way too much) — it's just that libjxl is not particularly frugal when it comes to memory usage, so the hack might be a placebo, you decide. It seems to help _somewhat_, presumably because forcing libvips to drop its objects at the end of each thumbnail operation reduces heap fragmentation. Using jemalloc via LD_PRELOAD helps with the rest — perhaps this is README material.

---

Plumbing progressive decoding through libvips would be a fun project, but not right now. A number of libvips loaders already support the notion of "decoding an image in a reduced resolution" and the thumbnailer already makes use of it where possible, so the infrastructure is there, but it needs a lot of plumbing and carefully reading jxl format spec to make it happen.

The actual no-decoding part (i.e. throw away HF groups) would probably require something more invasive if we want to do it within libvips and not turn copyparty into an underspecified implementation of half of ~~Common Lisp~~ the JXL bitstream spec :upside_down_face:

9001:
sweet! and thanks for the refactor/cleanup too :pray: :grin: 

> A number of libvips loaders already support the notion of "decoding an image in a reduced resolution" and the thumbnailer already makes use of it where possible

yeah, let's just wait for that to show up for jxl as well, no rush.

----

I've been looking into the memory usage trends you mentioned; the TLDR is I'm not seeing the same extreme values that you are, and I'm really curious why.

Granted, I haven't been using vips much since all my deployments are pil+ffmpeg, but this is what I'm seeing right now:

* not much difference between jpg and webp (true for both vips and pil), but jxl is heavy
* roughly the same peak memory usage when thumbnailing 100 vs 1000 jpeg/webp images (`94` vs `99` MiB) using vips
  * those numbers were with alpine in podman, using the musl allocator
  * by adding `LD_PRELOAD=/usr/lib/libmimalloc-secure.so.2` I get `876` vs `942` MiB(!)
* jpg/webp on my host os (archlinux) eats 357 MiB with vips, and 347 after adding your memory tweaks, so that's a Certified Good Change
  * but enabling jemalloc brings it up to 593 MiB
* jpg/webp with `--th-dec pil` to disable vips and use pillow...
  * the alpine-in-podman peaks are 461 MiB with mimalloc vs. 63 with musl
  * the system-archlinux peaks are 440 MiB with glibc vs. 652 with jemalloc
* and finally, jxl:
  * 877 MiB, archlinux / vips / glibc
  * 669 MiB, archlinux / vips / jemalloc
  * numbers for alpine in a bit; need to build the container first :p

and, just for the record, some notes on how I measured:

I'll admit my approach to measuring the alpine podman memory usage is pretty jank, but I *think* it's correct:

```bash
(sleep 1; free -k >/dev/shm/r1; curl -s '127.1:3923/?tar&j' >/dev/null; free -k >>/dev/shm/r1; podman stats --no-stream --no-reset >/dev/shm/r2; xdotool key ctrl+c; sleep 0.3; cat /dev/shm/r{2,1}; free -k) & rm -rf srv/pics/many/100/.hist/ ; /usr/bin/time podman run -e LD_PRELOAD=/usr/lib/libmimalloc-secure.so.2 --rm -it -p3923:3923 -v $PWD/srv/pics/many/100:/w localhost/copyparty-iv-amd64 -v /w::r --th-dec vips
```

the measurements with glibc vs jemalloc on archlinux are comparatively sane;

```bash
(sleep 0.5; curl -s '127.1:3923/?tar&j' >/dev/null; kill $(ps aux | awk '/[0-9]{2} python3[ ]-m copyparty/{print$2}')) & rm -rf srv/pics/many/.hist/ ; /usr/bin/time python3 -m copyparty -qv srv/pics/many::r --th-dec vips
```

my testcase is a folder with 100 images, 2870x1790 jpegs at q30 (filesize 255052 bytes), various crops of the same image:

```bash
for a in {0..9}; do for b in {0..9}; do convert ~/Downloads/2880x1800.png -quality 30 -crop 2870x1790+$a+$b $a$b.jpg; done; done
```

(I tried quality 80 as well, filesize around 1 MiB each pic, but no remarkable difference in memory usage)

PR #1282, opened 2026-02-07
GraalPy improvements

vgskye:
- Implements `ifaddr` on GraalPy by using Java's network interface APIs. (cursed!)
- Works around a partially broken `inet_pton` implementation when using GraalPy's Java POSIX backend

### DIFF
--- a/copyparty/util.py
+++ b/copyparty/util.py
@@ -270,6 +270,20 @@ try:
 
     socket.inet_pton(socket.AF_INET6, "::1")
     HAVE_IPV6 = True
+
+    if GRAAL:
+        try:
+            # --python.PosixModuleBackend=java throws OSError: illegal IP address
+            socket.inet_pton(socket.AF_INET, "127.0.0.1")
+        except:
+            _inet_pton = socket.inet_pton
+
+            def inet_pton(fam, ip):
+                if fam == socket.AF_INET:
+                    return socket.inet_aton(ip)
+                return _inet_pton(fam, ip)
+
+            socket.inet_pton = inet_pton
 except:
 
     def inet_pton(fam, ip):


9001:
I installed graalpy on an archlinux vm with `pyenv install -v graalpy-25.0.2` (which afaict is the recommended way), but the ifaddr change doesn't seem to work -- I got the following:

```
File "/tmp/pe-copyparty.1000.21160.0/copyparty/stolen/ifaddr/_graalvm.py", line 1, in <module>
  from java.net import NetworkInterface, Inet4Address, Inet6Address
KeyError: 'host symbol net.NetworkInterface is not defined or access has been denied'
```

also regarding the inet_pton fix, can you give an example of what this fixes, just so I can check too?

and should say I'm a fan of making the necessary changes to let graalpy use the essential features of copyparty, but perhaps we are going a bit too far with this -- is there a reason to use graalpy in favor of cpython or pypy? In a quick benchmark I saw graalpy was running at about half the speed of cpython, but also 10x the memory usage.

vgskye:
yeah, the default "native" distribution of graalpy does not ship with the full complement of java stdlib- you need the -jvm variant, which admittedly has horrible start-up time. this is also why it's surrounded with a try-catch.  
as for the inet_pton fix, passing `--python.PosixModuleBackend=java` to GraalPy should show the issue without the workaround.  
as for reasons to use GraalPy, it can make use of GraalVM Native Image to package copyparty and some optional dependencies into a single native executable, which is a fun trick- though, admittedly, not that useful since there are other ways to do it. the inet_pton change was actually found because GraalPy, when packaged with Native Image this way, uses the Java backend for virtual filesystem reasons and thus breaks without PRTY_NO_IPV6.

vgskye:
and yes, this means the graalpy ifaddr implementation works in very few cases; I can remove that if you'd like

9001:
alright, so the goal is to keep IPv6 support by keeping the original inet_pton for everything except IPv4 -- would this work? mainly because this way it would disengage if graal every fixes the bug, and also makes it compress a bit better;

```patch
diff --git a/copyparty/util.py b/copyparty/util.py
index 78ba922e..f27a5807 100644
--- a/copyparty/util.py
+++ b/copyparty/util.py
@@ -270,6 +270,20 @@ try:
 
     socket.inet_pton(socket.AF_INET6, "::1")
     HAVE_IPV6 = True
+
+    if GRAAL:
+        try:
+            # --python.PosixModuleBackend=java throws OSError: illegal IP address
+            socket.inet_pton(socket.AF_INET, "127.0.0.1")
+        except:
+            _inet_pton = socket.inet_pton
+
+            def inet_pton(fam, ip):
+                if fam == socket.AF_INET:
+                    return socket.inet_aton(ip)
+                return _inet_pton(fam, ip)
+
+            socket.inet_pton = inet_pton
 except:
 
     def inet_pton(fam, ip):
```

I built a native image, not sure I did it right but it ended up at 300 MiB and the performance wasn't exactly stellar :p so let's drop the ifaddr stuff -- however I bet the original ifaddr project would like it! https://github.com/ifaddr/ifaddr

and/or could also keep that change as a .patch in the repo in case anyone needs it

vgskye:
changes applied! yeah, GraalPy isn't exactly the fastest way to run copyparty, partly because its GIL seems to completely block multithreaded anything in my testing; I tried hackily patching GraalPy itself to remove the GIL(we have free-threaded python at home :3) and that significantly improved the performance.

as an aside, are your benchmarking scripts in the repo? `filehash.sh` was the only bench I could find and that showed Graal as way slower than CPython due to seemingly GIL issues

9001:
> changes applied!

sweet thx :grin: 

> as an aside, are your benchmarking scripts in the repo?

I keep reminding myself I should make a note somewhere, but the demotivating factor is "I should clean this up a bit first" :p

but basically I run two copyparties like:

```bash
taskset -c 0 python3 sfx1.py --no-ses -nw --wram -v /dev/shm/v::A -i ::,unix:/dev/shm/s1 -p 3921 -q
taskset -c 0 python3 sfx2.py --no-ses -nw --wram -v /dev/shm/v::A -i ::,unix:/dev/shm/s2 -p 3922 -q
```

and measure with:

```bash
q=?reset; for _ in 1 2 3 4; do for n in {1..10}; do for p in 1 1 2 2; do pp=$p; zp=$(printf "%0${p}d" $p); a=(); taskset -c 2 oha -wz500ms -c1 --output-format json --no-tui --ipv4 --insecure --unix-socket /dev/shm/s$p "${a[@]}" http://127.1:392$p/"$q" | grep requestsPerSec | sed -r $'s/.*: //;s/(\\...).*/\\1/;s/$/ \033[3'${pp}m$zp$'\033[0m/' | tee /dev/stderr; done; done | sort -n ; done 
```

* `q=?reset` is one of the cheapest GET-requests, the one I ran against graalpy was `q=?ls` and 20 tiny textfiles in `/dev/shm/v`
* `for p in 1 1 2 2` is the ports of the two copyparty instances, 392x
* `for _ in 1 2 3 4` is just for warmup, to get more stable measurements; last run is good
* `--unix-socket /dev/shm/s$p` is optional, didn't include it for this due to graalpy
* `taskset -c X` is removed when freethreaded or multiprocessing; helps for reducing noise otherwise
* `oha` is https://github.com/hatoo/oha

PR #1288, opened 2026-02-09
add check for alternative clipboard url on ctrl-c

icxes:
pretty self-explanatory innit
ui options will be added in the far future

### DIFF
--- a/copyparty/web/browser.js
+++ b/copyparty/web/browser.js
@@ -4622,7 +4622,7 @@ var fileman = (function () {
 		r.ccp = true;
 		r.clip = vps.slice(2);
 
-		var prefix = sread("clip_au") || location.origin;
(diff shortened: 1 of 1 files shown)


copyparty/web/browser.js
```diff
@@ -4622,8 +4622,9 @@ var fileman = (function () {
 		r.ccp = true;
 		r.clip = vps.slice(2);
 
+		var prefix = sread("clip_au") || location.origin;
```
```suggestion
		var prefix = sread("clip_uo") || location.origin;
```
"url-origin" because "au" makes it sound audio-related, wdyt?

icxes: yeah sounds good, it was "clip(board)_a(lt)u(rl)" in my mind but :+1:

9001:
:ok_hand:

PR #1296, opened 2026-02-14
Esperanto translation: silence skipping

slashdevslashurandom:
Some of the descriptions are guesses at this point (I assume "atk" functions like the "attack" value in an ADSR envelope, setting the amount of time during which the volume/speed changes)

### DIFF
--- a/copyparty/web/tl/epo.js
+++ b/copyparty/web/tl/epo.js
@@ -325,7 +325,7 @@ Ls.epo = {
 	"mt_ssvt": "laŭteca intervalo por silentsaltado (0-255)\">laŭt",
 	"mt_ssts": "komenca intervalo por silentsaltado (% de trakolongo)\">kom",
 	"mt_sste": "fina intervalo por silentsaltado (% de trakolongo)\">fin",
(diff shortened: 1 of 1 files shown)


9001:
> I assume "atk" functions like the "attack" value in an ADSR envelope

yup, that's the basic idea -- it's how quickly the volume and speed is changed when setting the new value, imagining there's a knob you can twist to adjust the volume it's how quickly you turn the knob :+1:

slashdevslashurandom:
Now that I think about it, a more common term for that, at least used in audio software aimed at listeners rather than creators, is "fade", or more specifically "fade-in" (gradual increase) or "fade-out" (decrease).

I assume the `atk` value is used for both (when there's a silence detected, the player first gradually changes the volume/speed to the skip values during `atk` seconds, goes past the silence, then gradually changes them back during another `atk` seconds)? In which case, a generic "fade" would function better as an English short label for the item.

9001:
yep, I'll swap out the wording in the english reference translation, feel free to update the epo meanwhile

slashdevslashurandom:
Esperanto doesn't have an unambiguous term for "fade" (as in audio effect, "dissolvo" seems to be used by some programs for the video effect), and since the `atk` menu item is used to specify the interval during which the volume/speed of playback is to change in context of skipping silence, I just picked "ŝanĝtempo" (literally "change-time") and replaced the "tempo" part with a stopwatch emoji for extra shortening. The tooltip message should provide enough clarification anyway.

9001:
sweet, thx! just in time for the next release :grin:

PR #1299, opened 2026-02-15
Support 4-digit octal values for `chmod_d` and `chmod_f`

new-sashok724:
Currently, the validation for the `chmod_d` and `chmod_f` volume flags only accepts 3-digit octal values. This prevents users from applying special permission bits (setuid, setgid, and the sticky bit), which are useful for shared group directories.

```
config-option 'chmod_d' must be a three-digit octal value such as [755] or [644] but the value was [2750]
```

This PR extends the validation to support 4-digit values and updates the help text to mention these special bits. [os.chmod](https://docs.python.org/3/library/os.html#os.chmod) already supports these bits, so no changes beyond that are needed.

\-\-\-


9001:
awesome, thanks :>

PR #1303, opened 2026-02-21
audioplayer: "fix" s-s error with high fade value (+extra)

### LINKED ISSUE #1302: Music player crashes with high ffwd values
### Describe the bug
When setting the ffwd on the skip silence feature to a value like 100, it brings you to the crash page

### To Reproduce
set ffwd to 100, and play some audio file in the copyparty webui
(tested on https://a.ocv.me/pub/demo/music/aivi%20%26%20surasshu%20-%20The%20Black%20Box/)
Image

### Screenshots
Image

icxes:
fixes #1302 (sort of)
- the current fade config value was kinda doing what ffwd was supposed to
  - solution: remove it. there isn't much point in having a fade config value as changing it doesn't make much of a difference except at extreme values (0.2 seems fine). having it is just extra confusion
- browsers stop playing audio at very low or very high playback rates (firefox does this at around 0.15 and 8.0 respectively based on my testing), so clamp the values between those. without this, the feature stops working above/below these values, as there is no audio to detect
- remove fade tooltip localization since it isn't needed anymore


9001:
:100:

PR #1312, opened 2026-02-25
tl: comprehensive overhaul and refinement of Chinese (chi.js) translation

lxdlam:
This PR provides a major update to the Chinese translation, moving away from literal/machine translations toward more standard and natural terminology used in modern Chinese software interfaces.

I didn't test every corner since I don't know how to invoke every feature, but I tried the major flows, which seem to have no issues.


9001:
thanks! appreciate the update, it's been a while since the last human intervention :>

i plan to look it over properly this weekend, but it looks good at first glance 👍

satgo1546:
Hi — I was a bit surprised when I figured out that someone had been working on exactly the same thing at exactly the same time as me!

I would (shamelessly) say that my translations were more comprehensive and overall better than yours :p

Nonetheless, I went over your changes and picked strings where I believed yours would be more legible.

You can see my version (now contains a part of your work) in #1316. Feel free to comment on things unclear/unnatural/...!

copyparty/web/tl/chi.js
```diff
@@ -244,28 +244,28 @@ Ls.chi = {
 
 	"cut_datechk": "除非启用「加速」按钮，否则没有效果$N$N略微减少 YOLO 因素；检查服务器上的文件时间戳是否与你的一致$N$N<em>理论上</em> 应该能捕捉到大多数未完成/损坏的上传，$N但不能替代之后禁用「加速」进行的验证\">日期检查",
 
-	"cut_u2sz": "每个上传块的大小（以 MiB 为单位）；较大的值跨大西洋传输效果更好。在非常不可靠的连接上尝试较小的值",
+	"cut_u2sz": "每个上传块的大小（以 MiB 为单位）；较大的值在“跨大西洋传输”效果更好。在非常不可靠的连接上尝试较小的值",
```
satgo1546, line 247: I'm not sure about this:

```json
	"cut_u2sz": "size (in MiB) of each upload chunk; big values fly better across the atlantic. Try low values on very unreliable connections",
```

@9001 is there anything special about atlantic? What about the pacific, or any long route across the globe?

We don't say atlantic very often, so I'm thinking of a rephrase.

lxdlam:
Oh it's great to find a colleague that also concern about the Chinese users!

I'll close this if you'll put more effort on this topic. Let's discuss in your pr.

copyparty/web/tl/chi.js
9001: a rephrase would be good; the original meaning is any long-distance transfer, generally between continents.

PR #1315, opened 2026-02-26
add optional checking for vulnerable versions

icxes:
if `vc-url` is set:
- reads ~/.config/vuln_advisory.json for vulnerability info if it is less than `vc-interval` old
- otherwise checks `vc-url` for a json containing vulnerability info as in https://api.github.com/repos/9001/copyparty/security-advisories
- if found, writes the response to ~/.config/vuln_advisory.json
- if the current version is vulnerable, lets the rest of the program know, if not, hits the bed for a day

or at least I hope that's what it does, I may have made some assumptions that may or may not be regarded, waiting warmly for review. i did test it and it seemed to work but i'm not really testerman so wouldn't be surprised if i overlooked something

### DIFF
--- a/contrib/systemd/copyparty.example.conf
+++ b/contrib/systemd/copyparty.example.conf
@@ -18,14 +18,14 @@
   # (note: enable compression by adding .xz at the end)
   q, lo: $LOGS_DIRECTORY/%Y-%m%d.log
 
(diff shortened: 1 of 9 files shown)


icxes:
i'm thinking the message might not be visible enough if it's only in the admin panel (at least i personally rarely see /?h), so this can probably stay until i work that out

9001:
thx o/

PR #1323, opened 2026-02-28
 Account for `rp-loc` in paths when `site` is set

new-sashok724:
Currently, when uploading a file via up2k, the returned upload URL path does not include `rp-loc` in the URL path if `site`/`up-site` is also set.

For example, if `rp-loc` is `copyparty`, the resulting URL is `https://domain.com/Volume/Folder/` instead of `https://domain.com/copyparty/Volume/Folder/`.

This issue occurs because `up_site`/`shr_site` is treated as a full URL prefix in the backend, but only as the domain/origin in the frontend. When `up_site`/`shr_site` is set, `rp_loc` is not included in the path. This PR fixes the issue by treating `up_site`/`shr_site` as the domain/origin everywhere, so that the full URL prefix is `up_site` + `rp_loc`.

\-\-\-

### DIFF
--- a/copyparty/httpcli.py
+++ b/copyparty/httpcli.py
@@ -2762,8 +2762,9 @@ class HttpCli(object):
         vpath = quotep(vpath)
 
         if self.args.up_site:
-            url = "%s%s%s" % (
+            url = "%s%s%s%s" % (
                 self.args.up_site,
+                self.args.RS,
                 vpath,
                 vsuf,
             )
@@ -3048,7 +3049,7 @@ class HttpCli(object):
                 raise Pebkac(500, t % zt)
             ret["purl"] = vp_req + ret["purl"][len(vp_vfs) :]
 
-        if self.is_vproxied and not self.args.up_site:
+        if self.is_vproxied:
             if "purl" in ret:
                 ret["purl"] = self.args.SR + ret["purl"]
 
@@ -3957,7 +3958,7 @@ class HttpCli(object):
 
             vpath = vjoin(upload_vpath, lfn)
             if self.args.up_site:
-                ah_url = j_url = self.args.up_site + quotep(vpath) + vsuf
+                ah_url = j_url = self.args.up_site + self.args.RS + quotep(vpath) + vsuf
                 rel_url = "/" + j_url.split("//", 1)[-1].split("/", 1)[-1]
             else:
                 ah_url = rel_url = "/%s%s%s" % (self.args.RS, quotep(vpath), vsuf)
@@ -6290,10 +6291,10 @@ class HttpCli(object):
 
         if self.args.shr_site:
             site = self.args.shr_site[:-1]
-        elif self.is_vproxied:
-            site = self.args.SR
         else:
             site = ""
+        if self.is_vproxied:
+            site += self.args.SR
 
         html = self.j2s(
             "shares",
@@ -6459,9 +6460,10 @@ class HttpCli(object):
 
         # NOTE: several clients (frontend, party-up) expect url at response[15:]
         if self.args.shr_site:
-            surl = "created share: %s%s%s/%s" % (
-                self.args.shr_site,
-                self.args.shr[1:],
+            surl = "created share: %s%s%s%s/%s" % (
+                self.args.shr_site[:-1],
+                self.args.SR,
+                self.args.shr,
                 skey,
                 fn,
             )


9001:
this change is not necessary in my own test setup; do you know why it appears to be necessary for you?

Screenshot_2026-02-28_16-29-29-or8

my copyparty commandline:  
`python3 -m copyparty -v srv/junk::A --rp-loc /foo --rproxy -1`

my nginx config: [etc-nginx-copyparty.conf.txt](https://github.com/user-attachments/files/25622556/etc-nginx-copyparty.conf.txt)

new-sashok724:
Interesting! :smile:
This seems to be an issue because I'm also using the `site` option -- if I remove it, it works. I'll check how it's used in the source and update the fix accordingly.
Btw, there seems to be another (minor) bug in your screenshot: `foo` (`rp-loc`) shouldn't appear in the queued filenames, right?

image

Here's my config:

<details>
<summary>Config</summary>

```Yaml
[global]
    # General
    name: Copyparty @ server.s724  # Server name (displayed topleft in browser and in mDNS).
    name-url: https://server.ps724/copyparty/  # URL for server name hyperlink (displayed topleft in browser).
    site: https://server.ps724/copyparty/  # Public URL to assume when creating links.
    hist: /srv/copyparty/hist/  # Where to store volume data (db, thumbs).

    # Logs
    q  # Quiet; disable most STDOUT messages.
    lo: /var/log/copyparty/copyparty.log  # Log file.
    flo: 2  # Log format for -lo; [1]=classic/colors, [2]=no-color.
    no-logflush  # Don't flush the logfile after each write.
    log-date: %Y-%m-%d  # Date-format (default is disabled; no date, just HH:MM:SS).

    # Network
    i: unix:660:copyparty:/run/copyparty/copyparty.sock  # IPs and/or unix-sockets to listen on.
    http-no-tcp  # Do not listen on TCP/IP for http/https; only listen on unix-domain-sockets.

    # Reverse proxy
    rproxy: 1  # Which ip to associate clients with; [0]=tcp, [1]=origin (first x-fwd).
    rp-loc: /copyparty  # If reverse-proxying on a location instead of a dedicated domain/subdomain, provide the base location here.
    no-crt  # Disable automatic certificate creation.
    http-only  # Disable SSL/TLS -- force plaintext.

    # Auth
    auth-ord: pw  # Controls auth precedence.
    ah-alg: argon2  # Account-pw hashing algorithm.
    usernames  # Require username and password for login; default is just password.
    dav-auth  # Force auth for all folders.
    dl-list: 1  # Who can see active downloads in the control panel? [0]=nobody, [1]=admins, [2]=everyone.
    ups-who: 1  # Who can see recent uploads on the ?ru page? [0]=nobody, [1]=admins, [2]=everyone.

    # Security
    xvol  # Never follow symlinks leaving the volume root, unless the link is into another volume where the user has similar access.
    vol-or-crash  # If a volume's folder does not exist on the HDD, then burst into flames.
    dont-ban: any  # Anyone at this access level or above will not get banned: [any]=disable-all-bans.
    cookie-lax  # Allow cookies from other domains.
    ver  # Show version on the control panel.

    # Upload
    dotpart  # Dotfile incomplete uploads, hiding them from clients unless -ed.
    chmod-f: 0640  # Unix file permissions to use when creating files.
    chmod-d: 0750  # Unix file permissions to use when creating directories.
    reflink  # Enable reflink-based dedup; will fallback on full copies when that is impossible (non-CoW filesystem).
    rm-partial  # Delete the .PARTIAL file when an unfinished upload expires after --snap-drop.

    # Transcoding
    no-bacode  # Disable batch audio transcoding by folder download (zip/tar).

    # View
    localtime  # Default to local timezone instead of UTC.
    theme: 1  # Default theme to use ([1] = classic light).
    ui-filesz: 5c  # Default filesize format.
    nsort  # Default-enable natural sort of filenames with leading numbers.
    sort: ext,href  # Default sort order, comma-separated column IDs, prefix with '-' for descending.
    au-vol: 100  # Default audio/video volume percent.

# Accounts
[accounts]
    sashok724: <REDACTED>

# Volumes
[/Storage/]
    /srv/copyparty/storage/

    # Volume accounts
    accs:
        A: sashok724

    # Volume flags
    flags:
        chmod_d: 2750

[/Samba/]
    /mnt/storage/samba/

    # Volume accounts
    accs:
        r.: sashok724

[/Syncthing/]
    /mnt/storage/syncthing/

    # Volume accounts
    accs:
        r.: sashok724
```

</details>

new-sashok724:
`rp_loc` is prepended in `HttpCli#handle_post_json` (`httpcli.py`), but only if `up_site` is not set. If I remove this condition, it works as on your screenshot.

It appears that `up_site` is treated as a full URL prefix in the backend but only as the domain/origin in the frontend. I think it would be better to treat `up_site` as the domain/origin everywhere, so that the full URL prefix is `up_site` + `rp_loc`.

I've updated the PR. Please let me know if you'd like me to fix it differently :)

9001:
> It appears that up_site is treated as a full URL prefix in the backend but only as the domain/origin in the frontend

ah, yes that was a mistake on my part -- the intention was that `up_site` would be the full URL prefix everywhere, in case someone wanted to use path-based proxying for the local domain but not for the external one -- but I now realize that this would not work reliably; it would have an edgecase regarding folders that happen to have the same name as the proxy-path, so your approach is better.

thanks!

PR #1335, opened 2026-03-04
fix(rotf): dupe file creation by stripping trailing slashes

### LINKED ISSUE #1334: Duplicate file creation when rotf has trailing slash
### To Reproduce

Drag and drop files from a file manager to browser

### Expected behavior

Only one copy of each file is uploaded

### Screenshots

https://github.com/user-attachments/assets/9e733577-5fba-4d54-9f4d-925faba5959c

### Server details

* **server OS / version:**
  * OS: `Arch Linux x86_64`
  * Kernel: `6.18.13-arch1-1`
  * Python version: `CPython v3.14.3`
  * GCC version: `gcc (GCC) 15.2.1 20260209`
  * Deps: `sqlite 3.51.2*1 | jinja 3.1.6 | pyftpd (None) | tftp (None) | miko (None)`
  * Copyparty version: `copyparty v1.20.10 "sftp is fine too" (2026-02-25)`
  * Docker version: `v29.2.1`
  * Docker image version: `copyparty/ac:1.20.10 linux/amd64`
* **how you're running copyparty:** Behavior persist across all 3 methods I tried.
  * `copyparty-sfx.py`: `python3 copyparty-sfx.py -c copypart.conf`
  * `copyparty` installed in a virtualenv via pip: `copyparty -c copyparty.conf`
  * `copyparty/ac` Docker image: Very bare-bones compose file, similar to the [example](https://github.com/9001/copyparty/blob/7c60adc69c977fb4b75e31c576131b1379498c3f/docs/examples/docker/basic-docker-compose/docker-compose.yml) in the repo
* **copyparty arguments and/or config-file:**

```yaml
# Docker one is slightly different regarding paths
[global]
  e2dsa
  e2ts
  z
  no-ansi
  lo: %Y-%m-%d.log

[/]
  .
  accs:
    r: *

[/public]
  ./public/
  accs:
    rw: *
  flags:
    rotf: %Y-%m-%d/
    # trailing slash at the end
```

### Client details

* browser version: `Mozilla Firefox 148.0`

### Server log

[2026-03-04.log](https://github.com/user-attachments/files/25729979/2026-03-04.log)

### Additional context

* Only an issue when using up2k
  * When `dedup` is in the config or is passed as an argument, the behavior still persists, but instead of a copy, a link is made
  * The behavior is consistent when `overwrite existing files on server?` is `never` or `unconditionally skip`
    * `never` creates a copy/link
    * `unconditionally skip` throws an error

NecRaul:
`rotf` is meant to be a directory abstraction, so it’s easy to assume that adding a trailing slash like:

```
%Y/%m/%d/
```
should be fine, just like writing a folder path with `/` at the end.

Right now, that trailing slash makes `copyparty` treat the same folder as two different paths, which can cause duplicate uploads (and confusion, in my case) with `up2k`. Stripping trailing slashes just makes the behavior consistent and prevents these surprises.

Fixes #1334

---

### DIFF
--- a/copyparty/authsrv.py
+++ b/copyparty/authsrv.py
@@ -190,12 +190,12 @@ class Lim(object):
             self.log_func("up-lim", msg, c)
 
     def set_rotf(self, fmt: str, tz: str) -> None:
-        self.rotf = fmt
+        self.rotf = fmt.rstrip("/\\")
         if tz != "UTC":
             from zoneinfo import ZoneInfo
 
             self.rotf_tz = ZoneInfo(tz)
-        r = re.escape(fmt).replace("%Y", "[0-9]{4}").replace("%j", "[0-9]{3}")
+        r = re.escape(self.rotf).replace("%Y", "[0-9]{4}").replace("%j", "[0-9]{3}")
         r = re.sub("%[mdHMSWU]", "[0-9]{2}", r)
         self.rot_re = re.compile("(^|/)" + r + "$")
 


9001:
nice, was expecting this to be trickier -- thx!

PR #1345, opened 2026-03-04
Started proof-reading and fixing the grammar of the README file

Artemis1-0:
This pull request is the first set of my updates to the readme.md file that fixes the grammar and wording in a few places. I decided to try the pr now since it may be a while before I get a chance to continue with my proof reading.
Have a good one!


9001:
thanks for opening the PR so early before you spent more time on it -- I do not think there is any point in trying to improve the readme, because many people have pointed out that it has fundamental structural issues which makes it impractical and inconvenient for most purposes.

a better approach would be to start afresh with a proper layout and organization, not in the form of a readme but instead a gitbook or more conventional manual.

as for the suggested changes specifically, I'm sorry to say but I would like to keep the current informal structure and layout -- I want this document to continue being a place to dump facts and knowledge in a rush without needing to think about structure and phrasing, which kills the joy of hacking on things (at least for me). Another reason to start from scratch really!

Artemis1-0:
That's fair to be honest. Would you be looking to use something like MkDocs?

PR #1351, opened 2026-03-08
feat: support .hidden file for dotfiles exclusion

NecRaul:
Add support for a per-directory `.hidden` file to exclude additional files/directories from listings.

- looks for a .hidden file in each directory being listed
- reads filenames from .hidden if it exists
- excludes any files or folders listed in `.hidden` from listings
- still hides normal dotfiles as before
- if no `.hidden` file exists, behavior is unchanged

### Example directory:

```text
/dir
├─ subdir/
│  ├─ lost+found/
│  └─ file2.txt
├─ lost+found/
├─ .secret
├─ .hidden  (contains "hidden.txt" and "lost+found")
├─ file1.txt
├─ hidden.txt
└─ lost+found.txt
```

### Resulting listing:

```text
/dir
├─ subdir/
│  ├─ lost+found/  (not hidden)
│  └─ file2.txt
├─ file1.txt
└─ lost+found.txt
```

---

### DIFF
--- a/copyparty/ftpd.py
+++ b/copyparty/ftpd.py
@@ -24,6 +24,7 @@ from .util import (
     ODict,
     Pebkac,
     exclude_dotfiles,
+    load_dothidden,
     fsenc,
     ipnorm,
     pybin,
@@ -348,7 +349,8 @@ class FtpFs(AbstractedFS):
             vfs_ls.extend(vfs_virt.keys())
 
             if self.uname not in vfs.axs.udot:
-                vfs_ls = exclude_dotfiles(vfs_ls)
+                dothidden = load_dothidden(fsroot)
+                vfs_ls = exclude_dotfiles(vfs_ls, dothidden)
 
             vfs_ls.sort()
             return vfs_ls
--- a/copyparty/ftpd.py
+++ b/copyparty/ftpd.py
@@ -24,7 +24,7 @@ from .util import (
     ODict,
     Pebkac,
     exclude_dotfiles,
-    load_dothidden,
+    exclude_dothidden,
     fsenc,
     ipnorm,
     pybin,
@@ -349,8 +349,10 @@ class FtpFs(AbstractedFS):
             vfs_ls.extend(vfs_virt.keys())
 
             if self.uname not in vfs.axs.udot:
-                dothidden = load_dothidden(fsroot)
-                vfs_ls = exclude_dotfiles(vfs_ls, dothidden)
+                if "dothidden" in vfs.flags and ".hidden" in [x[0] for x in vfs_ls]:
+                    vfs_ls = exclude_dothidden(vfs_ls, fsroot)
+                else:
+                    vfs_ls = exclude_dotfiles(vfs_ls)
 
             vfs_ls.sort()
             return vfs_ls
--- a/copyparty/httpcli.py
+++ b/copyparty/httpcli.py
@@ -80,6 +80,7 @@ from .util import (
     has_resource,
     hashcopy,
     hidedir,
+    load_dothidden,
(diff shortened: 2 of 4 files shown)


9001:
been digging into this a bit, and I'm not sure it's a good idea;

at best, this will be a basic filter that's applied when browsing around, and there are plenty of ways in which it will not apply -- files which are hidden by a `.hidden` will still appear in search-results, in the recent-uploads list, and the currently-ongoing-uploads/downloads lists, and if a folder is copied/moved then the `.hidden` is possibly/probably not carried over

if the intended purpose is just that -- a basic cosmetic filter similar to [--unlist](https://copyparty.eu/cli/#g-unlist) -- then we'll need to be hella clear about that in the documentation, so nobody thinks it's something it isn't

made some changes (hopefully improvements (i think)), now it's optional and default-disabled, and made it apply for download-as-zip/tar in addition to just listings, but even when it's not enabled it'll have a small performance impact on directory listings (1-2% maybe)... eh, probably fine [hidden1.patch](https://github.com/user-attachments/files/26163047/hidden1.patch)

and briefly tried to make it apply to the recent-uploads page before realizing it's mostly-impossible, since if the path is `qwe/rty/uio.mp3` and `rty` is hidden then it'd have to check the `.hidden` of each level going up which is just too expensive  [hidden2.patch](https://github.com/user-attachments/files/26163057/hidden2.patch)

NecRaul:
> if the intended purpose is just that -- a basic cosmetic filter similar to [--unlist](https://copyparty.eu/cli/#g-unlist) -- then we'll need to be hella clear about that in the documentation, so nobody thinks it's something it isn't

That was the behavior I intended and I should've been more clear in the PR message, so that's my bad. In fact, I'd like to make both the intended behavior and the intended use-case very clear here.

I modeled this `.hidden` behavior after default behavior on various GUI file managers (Dolphin, Nautilus, Thunar, PCManFM, Nemo, Caja - some of these I didn't even know about but they all respect `.hidden` as far as I could tell). That's why the PR title was `support` and not something else.

`.hidden` is per-directory, doesn't distinguish between file and folder, can't hide something that is nested and files/folders hidden by `.hidden` show up in search if they are indexed (Baloo Indexing for Dolphin for example).

The `.hidden` also has some weird behavior which I tried to carry over. For example, `qwe/` is not valid as none of the file managers I tried respected it. Each line can only contain one file or folder name and it should be exact. For example if the file name is `abc` while the line in `.hidden` is `abc ` (empty space at the end), `abc` file will not be hidden and vice-versa.

As for the intended use-case, it's for people who already use these GUI File Managers and have `.hidden` file in their workflow when using these file managers. As far as I'm aware, no applications generate `.hidden` file (or folder), which means, it's . That was the reason I made it default with no option to turn it off. Maybe that was an oversight on my part.

> made some changes (hopefully improvements (i think)), now it's optional and default-disabled, and made it apply for download-as-zip/tar in addition to just listings, but even when it's not enabled it'll have a small performance impact on directory listings (1-2% maybe)... eh, probably fine [hidden1.patch](https://github.com/user-attachments/files/26163047/hidden1.patch)

If you think it should be optional and default-disabled, that's completely fine by me. Regarding download-as-zip/tar thing you mentioned, either I'm misunderstanding what you're saying or I couldn't replicate the behavior - even while hidden, `.hidden`, files hidden by it and other dotfiles unrelated to `.hidden` were zipped.

9001:
sweet, yeah in that case it's all good -- I reverted the changes I did to make it... less cosmetic? stronger? something like that -- and just left the optimizations + making it default-disabled, gonna add some documentation ina bit -- thx o/

PR #1352, opened 2026-03-08
feat: adding handler for fail2ban

Lomaiin:
Added a handler to process 404 errors with fail2ban.

The handler respects the `--dont-ban` argument and, if necessary, will generate a special log line that will be intercepted by fail2ban.
Thanks to this, no one can spoof the log, and the fail2ban regex is lighter (and therefore faster).

---

### DIFF
--- a/bin/handlers/404-to-fail2ban.py
+++ b/bin/handlers/404-to-fail2ban.py
@@ -1,16 +1,27 @@
+# /!\ Warning: be careful, as webdav clients often generate a large number of 404 requets.
+
 # In your `jail.local`, add:
 # [copyparty]
 # enabled = true
+# logtimezone = UTC
 # logpath = /path/to/log/file # or keep the default value if you're using systemd
 
 # Create the `copyparty.conf` file in `filter.d` with the following:
 # [Definition]
-# failregex = ^ fail2ban\s*WARN: <ADDR>$
+# failregex = ^ <ADDR>$
 # ignoreregex =
-# datepattern = ^%%H:%%M:%%S\.%%f
+# datepattern = ^fail2ban: %%Y-%%m-%%dT%%H:%%M:%%S
 
 # First check `--dont-ban`, and if it doesn't match, log the line to be intercepted by fail2ban.
-def main(cli, vn, rem):
+from datetime import datetime, UTC
+def main(cli, vn="", rem=""):
+    now = datetime.now(UTC).isoformat()[:19]
+    msg = "\nfail2ban: %s %s"
+    if not vn and not rem:
+        # got called by --xban
+        cli["log"](msg % (now, cli["ip"]), 3)
+        return {"rc": 0}
+
     cond = cli.args.dont_ban
     if (
         cond == "any"
@@ -21,5 +32,5 @@ def main(cli, vn, rem):
     ):
         return ""
 
-    cli.log_func("fail2ban", f"{cli.ip}", 3)
+    cli.log(msg % (now, cli.ip), 3)
     return ""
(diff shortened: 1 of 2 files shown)


9001:
nice! happy to merge this as-is, but I have some ideas you might like :>

right now it only works with `--on404 404-to-fail2ban.py` but it wouldn't take much to also make it work for `--xban I,404-to-fail2ban.py` so it will also trigger for other types of sussy behavior;

```python
from datetime import datetime, UTC
def main(cli, vn="", rem=""):
    now = datetime.now(UTC).isoformat()[:19]
    msg = "\nfail2ban: %s %s"
    if not vn and not rem:
        # got called by --xban
        cli["log"](msg % (now, cli["ip"]), 3)
        return

    cond = cli.args.dont_ban
    if (
        cond == "any"
        or (cond == "auth" and cli.uname != "*")
        or (cond == "aa" and cli.avol)
        or (cond == "av" and cli.can_admin)
        or (cond == "rw" and cli.can_read and cli.can_write)
    ):
        return ""

    cli.log(msg % (now, cli.ip), 3)
    return ""
```

this also changes the message format slightly so it might be easier for fail2ban to read it; now it'll be just `fail2ban: 2026-03-21T20:37:31 127.0.0.1` on a new standalone line, so a regex like `^fail2ban:` would catch it

thing is, the syntax of copyparty's logfile is impractically different depending on several factors (`-q` or not, `--flo`, stdout or logfile) so was concerned about the example datepattern not hitting

I don't have a fail2ban setup to test any of this with, but what do you think?

9001:
oh and one thing to beware of;

many webdav clients (especially windows and macos) will send truckloads of requests that will 404 when they connect, because the very first thing that windows does when a new filesystem shows up is cheking whether it's a bluray disk... and lots of other bespoke things :p

so that's why copyparty is conservative with 404-bans out of the box

Lomaiin:
Ho yeah, I hadn't thought of `--xban`, that's a good idea !

> thing is, the syntax of copyparty's logfile is impractically different depending on several factors (`-q` or not, `--flo`, stdout or logfile) so was concerned about the example datepattern not hitting

The new message format is simpler, but I'll need to check for timezone correspondences, because the UTC time may not match the host's time zone, and I think that fail2ban needs to be aware of.

> this also changes the message format slightly so it might be easier for fail2ban to read it; now it'll be just `fail2ban: 2026-03-21T20:37:31 127.0.0.1` on a new standalone line, so a regex like `^fail2ban:` would catch it

My only concern of messages on a newline, is whether it’s possible for someone to tamper with a request by inserting a line break character to manipulate the line and trigger fail2ban. I tried using `%5C`, but without success. I'm asking to make sure it’s not possible.

I'll run some tests tomorrow and get back to you here.


And, for webdav, I do not use it for now, so I was not aware of this issue. By using `--xban`, copyparty handles the banning process, and thus keeps the conservative way. I'll add a warning comment to inform of this behavior.

Lomaiin:
Hi again !

As you suggested, I've took your modification to support `--xban` :)
I've changed a little bit the regex and the timezone configuration so that fail2ban continues to work.
I've changed the return type in case of `--xban` to prevent an AttributeError exception from being raised for `NoneType` on `hr`.

I've also tested it with `-q` and `--flo`:
- for `-q`, there is no output as intended -> if someone needs `-q` and fail2ban, the fix could be to use `print()` directly
- for `flo`, it works with both options

9001:
nice, thanks for adding this :>

it is possible to combine `-q` with `-lo` to disable stdout but still log to a textfile, and this changes the log syntax slightly (more optimized for logfile purposes) -- but the newline should ensure that it won't be an issue, so nothing to be done there :+1:

PR #1373, opened 2026-03-17
Removed color codes from log file

max-4-3:
Removed the annoying color escape codes from the log files.
`--no-ansi` does the same thing, but it also removes the color from console too :(

## Fix
patched the opened logfile's `write` method in `SvcHub` to first clean the message ( i.e. strip any ansii escape code using re ) and then log that to the file.
The Regex was taken from [R](https://regex101.com/r/96ZckU/1)

NOTE: I also add some code in `copyparty/__main__.py` that will do the cleanup using `logging` capabilities. just comment out and remove some local imports

## Before
image

## After
image

## Extras
Also added a `colors.txt` file containing all color codes used in `copyparty`  
Also added a grep cmd for retriving the above ;) [Only used in linux, but might work on modern verrsion]

This PR complies with the **[DCO](https://developercertificate.org/)**


9001:
this should already be covered by the `-flo` option; https://copyparty.eu/cli/#g-flo -- does your usecase require something else?

the colors being written to the logfile by default is intentional, because they display correctly when the logfile is viewed with for exampel `less -R` or with copyparty itself (the textfile viewer);

Screenshot_20260317-142141~2

Screenshot_20260317-142736~2

and there is the problem that simply stripping the colorcodes from the default format will lose the severity labeling so it becomes impossible to tell if something is a warning or error

9001:
ah dangit, that doesnt apply unless you also use -q ... man

that's a bug, should be fine if that is fixed right?

max-4-3:
> ah dangit, that doesnt apply unless you also use -q ... man

what do u mean?
-q just silents the main output, which is kinda not gud ;-;

max-4-3:
> this should already be covered by the `-flo` option

i didn't knew about that tbh, but yeah as you mentioned its kinda buggie :3

max-4-3:
Slapping an `if` check and woila a fix xD

9001:
ah but there's still this issue though,

> and there is the problem that simply stripping the colorcodes from the default format will lose the severity labeling so it becomes impossible to tell if something is a warning or error

 `-q` had the functionality that it would also reformat the loglines to preserve that info, as did `--no-ansi`

I think I found a way to fix this without losing too much performance -- 8c6d8a3c22ac33b8b86be8dc15c7bca2cfe41f75 so this should be good,

* beta SFX and whl: https://copyparty.eu/beta/
* beta docker image: `docker/ac:beta` on [dockerhub](https://hub.docker.com/layers/copyparty/ac/beta/images)

max-4-3:
oh i didn't take account for log level dumb me ;-;
btw u check for nullabilty of `self.logf` at `:1734` but didn't check for it in `:1782`? 
now we have 3 log fn xD

9001:
yeah -- it should be fine because that function only gets engaged when -lo is set, so the logfile definitely exists 👍

max-4-3:
oh thats cool
btw now what to do with this branch?

PR #1387, opened 2026-03-24
Add gentoo packaging

mid-kid:


9001:
nice, thanks for packaging it!

the readme ToC is autogenerated based on the first line in each section so tweaked it slightly to make it work o/

mid-kid:
Thanks!

Issue #1033, opened 2025-11-24
Folders made with Config file have inaccurate file sizes


vmattphillips:
different from https://github.com/9001/copyparty/issues/256
Im having the same issue but im running this directly from the python file

```
[global]
  lo: ./logs/cpp-%Y-%m%d.xz  # log to c:\users\you\logs\
  e2dsa, e2ts, z  
  p: 443  
  theme: 2  
  lang: eng
  usernames
  ah-alg: scrypt,13,2,8,4,32
  xff-src: lan
  xff-hdr: x-forwarded-for
  rproxy: -1

[accounts]               
  #Hashed passwords
  owner: +hash
  UserA: +hash
  UserB: +hash

[/] 
  e:\Copyparty\Drive       
  accs:           
    rwmda.: owner
    r: UserA, UserB, 
    #g: *       
 folders
    flags:
      e2d
      d2t
      fk: 4

[/PublicSharing]
  e:\Copyparty\Drive\PublicSharing
  accs:
    rwmda.: owner 
    wG: UserA, UserB
    g: *        
folders
    flags:
      e2d
      d2t
      fk: 4
    
[/UserA]               
  e:\Copyparty\Drive\UserA          
  accs:          
    rwmda.: owner
    rwd: UserA
    g: *        
 folders
    flags:
      e2d
      d2t
      fk: 4

[/UserB]              
  e:\Copyparty\Drive\UserB         
  accs:           
    rwmda.: owner 
    rwd: UserB
    g: *        
    flags:
      e2d
      d2t
      fk: 4

```

If I make folders at root, or within these config folders, they report the correct size
but these specific folders I set up with permissions in teh config show either 0 size or 4096, and deleting files below 4096 size does not update it either.

mircey:
i have the same issue: any volumes mounted via the config file lack real sizes. they show `4 KB` instead. all folders within these volumes however do list their sizes correctly.

9001:
volumes don't display with correct size when viewed from the parent folder; it's a limitation of how the sizes are collected for display, and fixing it will most likely be a performance hit, so it's not been a priority

vmattphillips:
Contributing to an open source project I actually use has been on my bucket list, would you be opposed to some optional flag that did this if I can dig into the repo and figure out how it all works?

I say optional flag mostly because I recognize it may not be worth a potential performance hit, even though admittedly from what I've seen using the app and reading the config I'm not 100% sure this is how you intended flags to work lol

PR #1390, opened 2026-03-24
Adding new flag for optional calculation of Volumes created at the root

vmattphillips:
Adding this as a potential fix for https://github.com/9001/copyparty/issues/1033

The block of code I changed in httpcli should now query into the volumes at the root and fetch their sizes, this shouldn't effect anything else, as its blocked behind an optional flag for "deep" searching the directory sizes. (deep-dirsz)

before it would pass over since the database excludes the config defined volumes from indexing, this just adds an extra check if the result was falsey / none, AND if the user actually wants to see the sizes

(I've never made a pr to a repo that wasn't just me and my friends / classmates in college sorry if something is wrong)


9001:
I tried having it precompute the sizes during rescan and store the sizes back into the database, but quickly converged on your approach since it's cleaner -- just tweaked it a bit to also handle nested subvolumes and now it should be all good, thx for the nudge o/

PR #1407, opened 2026-03-30
add autogrid

icxes:
adds autogrid as suggested in issue 1384
- switches (to/out of) grid view automatically if (over/under) ga_thresh% of files have a file extension that is in ga_ext
- always uses grid view if grid view is also enabled
- configs for enabling and setting ga_thresh and ga_ext


9001:
thx! simplfiied it a bit to use the existing regex for pics/vids, that's probably good enough -- can make it configurable if that turns out necessary o/

PR #1410, opened 2026-03-31
Add automatic browser locale detection using `--glang`

stackxp:
This adds an automatic browser locale detection that can be enabled with the `--glang` flag. Locales are detected using the `navigator.languages` field in JS. If it can't find a fitting language, it falls back to the default one specified by `--lang`.


9001:
thx!

PR #1419, opened 2026-04-04
Add script + pandoc template to turn README.md into docs site

Scotsguy:
[pandoc](https://pandoc.org/) is a very versatile tool for converting between document formats. one of the output formats it supports is "chunkedhtml", which consists of multiple html files, one for each top-level heading of the input.

inspired by [this tangled blogpost](https://blog.tangled.org/docs/), we can use pandoc and a custom template that moves the table of contents into a sidebar, we can turn the README.md into a quite passable docs site without rewriting everything:

Screen Shot 2026-04-04 at 21 37 31

TODO:
- mobile support
  - <details>Screen Shot 2026-04-04 at 21 38 22 </details>
  - tangled do the TOC with a popover for a JS-free solution, seems to work well
- ~~CSS for tables~~ done
- (maybe a later PR) headings don't have anchor links in the main text, only in the TOC. this needs [a lua hook](https://github.com/jgm/pandoc-website/blob/45c9c244c88c5f1c77a9af920f6e796ac858ba02/tools/anchor-links.lua).

needs discussion:
- I have no idea what I'm doing wrt. CSS, I would appreciate a review
  - ~~in particular, the hack for centering images while also letting them be wider than the text doesn't affect the scroll bar, so you can scroll to the right into nothingness~~ fixed (worked around?)
- relative links that used to point into the file tree (in particular, docs/) now point into nothingness. I'm not sure if it's a better idea to make these absolute links or also host docs/ or something else
  - perhaps replace the giant README.md with a normal-sized readme and move the current readme into docs/. that way we wouldn't have to worry about the links still working on github.
    - for ctrl+f purposes, tangled also assembles one giant single-page version of their docs.

### DIFF
--- a/scripts/docs/assemble-docs.sh
+++ b/scripts/docs/assemble-docs.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env sh
+
+rm ./out -r
+
+pandoc ../../README.md \
+  --output=out/ \
+  --from=gfm \
+  --to=chunkedhtml \
+  --standalone \
+  --table-of-contents \
+  --toc-depth=3 \
+  --css=./docs.css \
+  --chunk-template="%i.html" \
+  --template ./template.html \
+  --metadata title="readme" \
+  --title-prefix="copyparty"
+
+cp ./docs.css out/
--- a/scripts/docs/docs.css
+++ b/scripts/docs/docs.css
@@ -1,0 +1,135 @@
+html {
+	color: #ccc;
+	background: #222;
+	font-family: sans-serif;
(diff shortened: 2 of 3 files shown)


oisin404:
This looks amazing! Would be great so see something like this!

Scotsguy:
I'm unfortunately not going to have much time to work this the next few weeks; the main blocker currently is that the generated website is unusable on mobile. If anyone would like to contribute their web development knowledge I would be grateful, otherwise I'll see when I can spare some time.

PR #1425, opened 2026-04-08
Greater compatibility for *BSD

chilledfrogs:
There are references to `lsblk` (from util-linux) and /proc that don't exist on *BSD, so this expands on some macOS-specific code to compensate for that

I did my best to follow the existing style and structure of the code, however I might have missed something (and also given that `_extdevs_nix()` doesn't show IPv6 addresses by default that might need to be reworked in the future anyway)

Testing from OpenBSD and FreeBSD users more than welcome, I have thus far tested this briefly on my OpenBSD box

This also lays better groundwork for packaging on OpenBSD and FreeBSD which I will get to afterwards

### DIFF
--- a/copyparty/__init__.py
+++ b/copyparty/__init__.py
@@ -44,6 +44,14 @@ ANYWIN = WINDOWS or sys.platform in ["msys", "cygwin"]
 
 MACOS = platform.system() == "Darwin"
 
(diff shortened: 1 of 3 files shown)


9001:
nice! looks good, thx :>

PR #1428, opened 2026-04-09
make `.txt` the default extension for `text/plain`

### LINKED ISSUE #1427: `--magic` selects `.ssa` as extension for plain text nameless uploads
### Describe the bug
When enabling `--magic` and uploading a nameless file with plain text contents, the guessed file extension is not the expected `.txt` but `.ssa`.

I couldn't figure it out 100% but I strongly suspect it has something to do with the line `text ass=plain ssa=plain` [here](https://github.com/9001/copyparty/blob/ede692925edfc6fca7a52f483414d543e9303237/copyparty/util.py#L500). This ends up in `SAFE_MIMES` but the Magician class actually only seems to refer to `MIMES` for turning a type into an extension, so that's still a missing link for me. It smells a lot like a reverse lookup for `plain` that returns `ssa` when it should return `txt` because the aforementioned line comes _after_ the one with `txt=plain` so it takes precedence.

### Server details (if you are using docker/podman)
* **server OS / version:** v1.20.13
* **how you're running copyparty:** podman
* **docker image:** min (with py3-magic installed on top of it)

### Additional context
Probably not relevant, but the `py3-magic` package is at version 0.4.27-r3 in my container.

Scotsguy:
Closes #1427 

`MIMES.items()` iterates in insertion order, so the last-inserted entry had priority, meaning that `text/plain` got `ssa` as default extension.

### DIFF
--- a/copyparty/util.py
+++ b/copyparty/util.py
@@ -491,13 +491,12 @@ font woff woff2 otf ttf
         for v in vs.strip().split():
             MIMES[v] = "{}/{}".format(k, v)
 
-    for ln in """text md=plain txt=plain js=javascript
+    for ln in """text md=plain js=javascript ass=plain ssa=plain txt=plain
 application 7z=x-7z-compressed tar=x-tar bz2=x-bzip2 gz=gzip rar=x-rar-compressed zst=zstd xz=x-xz lz=lzip cpio=x-cpio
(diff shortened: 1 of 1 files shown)


9001:
nice! thx o/

PR #1439, opened 2026-04-15
Better ipv6 ratelimiting logic

SnowSquire:
This is a simple change that also allows for configuration of the truncated prefix length later. I changed it to /56 because that is the recommended prefix length for residential networks.


9001:
nice, and it's even a bit faster now :> thx!

PR #1458, opened 2026-05-01
Add macOS related information

ilotoki0804:
This PR adds macOS-related information to the README and devnotes.

* It appears that assigning ports 80 and 443 works without requiring root permissions, at least on macOS 26 Tahoe what I tested, so I updated the README accordingly.
* Adds information on how to install the GNU utilities required to build the sfx on macOS.

This PR complies with the DCO; [developercertificate.org](https://developercertificate.org/)


9001:
thx!

PR #1479, opened 2026-05-13
Add more dependencies to rc.d script.

Kansattica:
On my FreeBSD machine, copyparty wasn't starting with the computer even after being enabled. Turns out it was being started first, before all the filesystems and such showed up. This fixed it real nice.


9001:
thx!

PR #1482, opened 2026-05-15
invalidate get-only shares when creator no longer exists

### LINKED ISSUE #1480: get-only shares still active if user removed
if a user creates a get-only share, and the server-admin later removes that user, the share remains active

this behavior is unlike read- and/or write-shares, which are invalidated in that case, and they should behave the same way for consistency

initially reported by @celinke97

Scotsguy:
Fixes #1480

I noticed while making this fix: shares invalidated in this way don't actually get removed from the db. is that intended?

### DIFF
--- a/copyparty/authsrv.py
+++ b/copyparty/authsrv.py
@@ -3113,7 +3113,7 @@ class AuthSrv(object):
 
                 try:
                     s_vfs, s_rem = vfs.get(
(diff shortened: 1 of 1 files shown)


9001:
thx! and yes, it's intended in case the user was only removed temporarily (perhaps by mistake), so their shares reappear when the user is reinstated -- let me know if you can think of any reason that might be a bad idea :>

PR #1485, opened 2026-05-15
auth: read (and admin) implies get

### LINKED ISSUE #1483: Users with read permission do not count as having get permission
This prevents e.g. creating a share with get permission by users who have read permission, unless they are explicitly also granted get permission.

Scotsguy:
also, hide redundant get permission in webui

Closes #1483

hopefully this fix is at the correct layer; I didn't do a thorough search for things that may get affected by the permission alias, but making shares with get permission works when the creator has just read permission, and at least the webui won't get cluttered because of it

### DIFF
--- a/copyparty/authsrv.py
+++ b/copyparty/authsrv.py
@@ -1663,7 +1663,8 @@ class AuthSrv(object):
             for alias, mapping in [
                 ("h", "gh"),
                 ("G", "gG"),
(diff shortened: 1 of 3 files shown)


9001:
looks good at first glance, thx! I'll merge it when I've had a chance to check properly, might be a while but definitely before next ver :>

Issue #1486, opened 2026-05-15
Spectrograms in grid view: add adrgument to show frequency in log scale instead of linear


9hax:
I use copyparty to browse and manage my music library. 
I also love the spectrogram view to get an idea of how the song might be energy-wise.
Thus, I am somewhat frustrated that the spectrogram views are generated in linear scale.

**Describe the idea / solution you'd like**
I would love a toggle that i can set in the config or as an argument that allows ffmpeg to be called with the fscale:log toggle to generate spectrogram images with log frequency scale.

**Describe any alternatives you've considered**
downloading the file and analyzing it offline is an alternative, but honestly doesnt really compare to just seeing the spectrogram in the preview, right?

**Additional context**
I am honestly just writing this issue to reference it in a pull request that i will open right after this. 
I have already implemented this change. 

Left is a spectrogram with the linear frequency scale, right is one with the log scale. 
You can see that while most information on the left is lost and only high frequency info can be seen, the right side shows the frequencies clearly.

![](https://files.9hax.net/copyparty-log-frequency-scale.png)

PR #1487, opened 2026-05-15
feat: Add option to render spectrograms with log frequency scale

9hax:
This implements the feature described in #1486 .

It adds the argument ``th-spec-fl`` (thumbnails -> spectrogram -> frequency log scale) (please advise on a better name, this is what i decided on at 1am).
Adding this argument adds ``:fscale=log`` to the ffmpeg ``showspectrumpic`` filter, making it use a logarithmic scale instead of linear.

I tested this on my dev machine and it's now running in my prod deployment on my media server and seems to work well.

If spectrum pics were used previously, you have to clear the cache or non-linear pics will stick around.
Is there a fix for this, like invalidating all generated thumbnails if the arg was changed from the last run? 
Couldn't find anything like that in the code base yet.

Thank you for this awesome project, I hope this is good enough for a first contribution.

### DIFF
--- a/copyparty/__main__.py
+++ b/copyparty/__main__.py
@@ -1769,6 +1769,7 @@ def add_thumbnail(ap):
     ap2.add_argument("--th-pre-rl", metavar="SEC", type=int, default=30, help="while pregen is running, ratelimit the thumbnailer logger to one message every \033[33mSEC\033[0m seconds (only works with \033[33m-j1\033[0m); set 0 to disable ratelimit")
     ap2.add_argument("--th-covers", metavar="N,N", type=u, default="folder.png,folder.jpg,cover.png,cover.jpg", help="folder thumbnails to stat/look for; enabling \033[33m-e2d\033[0m will make these case-insensitive, and try them as dotfiles (.folder.jpg), and also automatically select thumbnails for all folders that contain pics, even if none match this pattern")
     ap2.add_argument("--th-spec-p", metavar="N", type=u, default=1, help="for music, do spectrograms or embedded coverart? [\033[32m0\033[0m]=only-art, [\033[32m1\033[0m]=prefer-art, [\033[32m2\033[0m]=only-spec")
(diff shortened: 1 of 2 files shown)


9001:
looks good!

if you want to make it automatically expire the thumbnail-cache when that option is changed, then this would do that:

```patch
diff --git a/copyparty/th_srv.py b/copyparty/th_srv.py
index 397f055d..e953cafe 100644
--- a/copyparty/th_srv.py
+++ b/copyparty/th_srv.py
@@ -465,10 +465,15 @@ class ThumbSrv(object):
     def volcfgi(self, vn: VFS) -> str:
         ret = []
         zs = "th_dec th_no_webp th_no_jpg"
         for zs in zs.split(" "):
             ret.append("%s(%s)\n" % (zs, getattr(self.args, zs)))
+        zs = "th_spec_fl"
+        for zs in zs.split(" "):
+            v = getattr(self.args, zs)
+            if v:
+                ret.append("%s(%s)\n" % (zs, v))
         zs = "th_qv th_qvx thsize th_spec_p convt"
         for zs in zs.split(" "):
             ret.append("%s(%s)\n" % (zs, vn.flags.get(zs)))
         return "".join(ret)
 
```

technically it would have been sufficient to just add `th_spec_fl` to the existing list of `th_dec th_no_webp th_no_jpg`, but that would discard everyone's cache for no real reason when they upgrade to the next version, so making it conditional like this instead avoids that :+1:

9hax:
I tried to implement your suggested change, but in my testing, the thumbnails are not invalidated when the argument is toggled.
I tried setting the timeout for the thumbnail cleaner to run every 60 seconds, but it doesn't seem to clean up the thumbnails even then.
Sadly, I don't know how to debug this.

9001:
huh, was sure that would be the only necessary change to make that work -- sorry for the trouble, I'll get it added

9hax:
nevermind... my browser had a bug for a short time which caused it to cache images too greedily. 
your suggested code works, i'm packaging it up in a nice commit for merging :3

9001:
sweet, thx o/

Issue #423, opened 2025-08-02
support for office files (collabora/openoffice)


bjornclauw:
**Describe the idea / solution you'd like**
It would be nice if we could open and edit office files in copyparty in the the browser.

icxes:
I think this might be a bit out of scope for the software

sharicov:
just like filestash does

erikboberg:
I would love this feature too!

icxes:
@9001

9001:
initial support is now in latest release; no documentation yet because i'm unfamiliar with optimal configuration of collabora online and don't want to give bad/dangerous advice #1575 , and onlyoffice is probably not yet supported #1574

PR #1495, opened 2026-05-23
WOPI integration

brandon-doornbos:
This PR adds WOPI integration to Copyparty, mainly from a Collabora Online perspective (hit me up if you tested other impls that didn't work). See below how it should work.

[Screencast_20260523_230350.webm](https://github.com/user-attachments/assets/ae087e2a-f3f5-4bf4-9421-2aee5eb6bb5a)

Adds an option to enable WOPI, to specify the WOPI client (e.g. self-hosted Collabora Online) and the session TTL. The client is by default set to the EU demo server of Collabora themselves, which you can use if you access your Copyparty through a publicly accessible URL (so Collabora can actually get your file).

WOPI expects a file-id and access-token for each session, we implement this through the Python Secrets API, generating 64 byte 'cryptographically strong' tokens for both. These then get stored in a dictionary in the http server to re-authenticate only that specific combination of file-id and access-token during the session. The tokens are removed from the dictionary after a configurable (default 30 min.) TTL.

(Future) TODOs:
- [Extra WOPI methods/options?](https://sdk.collaboraonline.com/docs/advanced_integration.html#checkfileinfo-response-properties)
- More config options (extensions to allow WOPI for, etc.)?
- Granular access control (we now just bail if you can't edit, but WOPI has stuff for only viewing etc.)


icxes:
related issue #423

brandon-doornbos:
Since this adds another way of authenticating, if someone else could check my logic would be nice :). Let me know if I need to change things or if this isn't appropriate for the project.

If it's close I'll start on documentation in the readme.

Demo WOPI client btw: https://demo.eu.collaboraonline.com

i0nicle:
would b awesome if this got merged B) would replace my seafile + collabora setup

9001:
haven't had a chance to look at this yet and will probably be a while still, but wanted to mention something right away --

copyparty does not and will never have support for locking files or folders, will never any sort of [file locking](https://en.wikipedia.org/wiki/File_locking) at all -- this is an intentinoal choice because the complications and safety implications of having to ensure that such locks behave correctly is not something I want to maintain or guarantee

so if that could be a problem then maybe this is a bad idea , I just recall someone saying wopi would expect correct locking behavior

brandon-doornbos:
Thanks for the heads-up, I'll put it in the documentation for other implementations. For Collabora anyway, [this actually has handling on their side](https://sdk.collaboraonline.com/docs/How_to_integrate.html#further-differences-to-wopi), although I'll look at properly handling it from our side. (edit, should be done, see [8304388](https://github.com/9001/copyparty/pull/1495/commits/8304388d299de608cbb87b6bfb0e46447af4f8ce))

brandon-doornbos:
I think this is pretty ok now, would someone be able to review it?

Have been using it myself the last few weeks without problems.

samchouse:
I'm not sure if the issue is on my side but OnlyOffice can't edit the document while Collabora can. In OnlyOffice, I get sent into a view only mode. Do you have any suggestions?

9001:
on that note, is this an accurate list of wopi clients worth testing with? or should some be added/removed?

* collabora online
* onlyoffice
* libreoffice online

9001:
at a quick glance (have not tried running the code in this PR yet),

using `ET.fromstring` is fundamentally unsafe, although possibly okay for this specific usecase. Regardless, copyparty has `dxml.parse_xml` which is a safe and functionally-identical alternative to `ET.fromstring`, so please use that instead. It is best to `from .dxml import parse_xml` next to the code that needs it, so it is not imported unnecessarily. If you run into issues with dxml's parser and it worked with raw ET.fromstring then that would imply the client is doing something chaotic-neutral, so let's hope not

`self.conn.hsrv.wopi_files` is okay as long as the server is running with `-j1` (default); if someone enables multiprocessing then this will break (clients are loadbalanced between hsrv's). That's fine, setting `-j` to anything other than 1 is almost always a mistake, almost always leading to reduced performance. Handling this properly is not worth it. Can just detect this misconfiguration and warn on startup:

```patch
diff --git a/copyparty/svchub.py b/copyparty/svchub.py
index d63b6308..50339d34 100644
--- a/copyparty/svchub.py
+++ b/copyparty/svchub.py
@@ -1109,3 +1109,3 @@ class SvcHub(object):
 
-        if not self.args.no_dav:
+        if self.args.wopi or not self.args.no_dav:
             from .dxml import DXML_OK
@@ -1113,6 +1113,11 @@ class SvcHub(object):
             if not DXML_OK:
-                if not self.args.no_dav:
-                    self.args.no_dav = True
-                    t = "WARNING:\nDisabling WebDAV support because dxml selftest failed. Please report this bug;\n%s\n...and include the following information in the bug-report:\n%s | expat %s\n"
-                    self.log("root", t % (URL_BUG, VERSIONS, expat_ver()), 1)
+                self.args.wopi = False
+                self.args.no_dav = True
+                t = "WARNING:\nDisabling WebDAV and WOPI support because dxml selftest failed. Please report this bug;\n%s\n...and include the following information in the bug-report:\n%s | expat %s\n"
+                self.log("root", t % (URL_BUG, VERSIONS, expat_ver()), 1)
+
+        if self.args.wopi and self.args.j != 1:
+            self.args.wopi = False
+            t = "WARNING: Disabling --wopi because -j is not 1 (the default and recommended -j value)"
+            self.log("root", t, 1)
 
```

`secrets` did not exist before python 3.6, but most of copyparty supports 2.7 -- if we want to use secrets, then can move `import secrets` into the function where it is used

haven't looked at `tx_wopi_api` / `tx_wopi` / `rx_wopi` properly yet, can't really brain these days but at least thats something :+1:

brandon-doornbos:
> I'm not sure if the issue is on my side but OnlyOffice can't edit the document while Collabora can. In OnlyOffice, I get sent into a view only mode. Do you have any suggestions?

Ah I looked into it and apparently OnlyOffice does that when it can't get a lock, since we will never support locking this is intended behaviour sadly. I'll mention in the documentation that it doesn't support editing in OnlyOffice due to this.

brandon-doornbos:
> on that note, is this an accurate list of wopi clients worth testing with? or should some be added/removed?
> 
>     * collabora online
> 
>     * onlyoffice
> 
>     * libreoffice online

The previous comment handles OnlyOffice, although I'll still try to keep it in mind for viewing.

As far as I know, Libreoffice don't have an online version themselves, Collabora is forked from Libreoffice to run online. Please correct me if I'm wrong though.

9001:
there's some cleanup work remaining, i'll make  afew more commits and some issues to track it

9001:
thx for all the work figuring out the protocol and all the other stuff! with some luck you guys might have an idea about the remaining things too :>

EDIT: beta sfx and whl available here (no docker images yet): https://copyparty.eu/beta/wopi/

brandon-doornbos:
Thank you for the revision and effort! I just started a new job so I'll have a bit less time, but I'll try to work on both issues when I can.

The idea for fake locking seems plausible, I was thinking about the same but didn't mention it since I didn't expect it to be okay. However since it's onlyoffice client-side anyway, it'll probably be fine.

9001:
awesome, good luck in the new job! and no rush at all 👍

PR #1506, opened 2026-06-02
add upload hotkeys: n & shift+n

juodumas:
Upload process before: 1) click up2k icon; 2) click upload; 3) click files or folder in modal
Upload process after: press N
Or do you all drag & drop?

Hotkey is `N` because it is opposite of `Y` and can also stand for `(N)ew`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -687,6 +687,8 @@ the browser has the following hotkeys  (always qwerty)
 * `ctrl-C` copy selected files/folders to clipboard
 * `ctrl-V` paste (move/copy)
 * `Y` download selected files
+* `N` upload file(s)
+* `shift-N` upload folder
 * `F2` [rename](#batch-rename) selected file/folder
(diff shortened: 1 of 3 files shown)


9001:
a huge inconvenience with initiating an upload by clicking the upload-button (and/or with this hotkey) is that it initiates file selection using a more restricted API than a drag-drop would;

the webbrowser has security features which prevents the javascript from recursing into a selected folder if it comes from the click-to-upload action,

so the only way to upload a folder and all its subfolders is by dragdrop; can't be done with the button

so I'm not really sure if this is worth it , but i'll keep it open for now -- figured i'd say something at least

juodumas:
Thanks. Then let's add only the file upload shortcut and skip the folder upload 
shortcut?

I don't use drag & drop uploads and three mouse clicks to upload is a pain.

PR #1511, opened 2026-06-05
Add contrib example for traefik and docker

tildezero:
The included traefik example uses the traefik yml file, and I know a lot of people (me included) use Docker labels to configure traefik, so I thought this would be helpful.


9001:
thx!

PR #1528, opened 2026-06-16
fix: improve directory sorting

### LINKED ISSUE #1119: Unexpected directory sorting order
### Describe the bug
If you have 2 directories: `a` and `a b`, `a b` is sorted before `a` when sorted by name ascending. Curiously, this only applies to directories: if they are files `a` and `a b`, `a` is correctly sorted before `a b`.

### Expected behavior
Sorting order for directories to behave like files.

NecRaul:
Ignores the trailing slash, directory key or any other type of query before sorting.

Makes the directory sorting consistent with the filesystem and file sorting behavior in copyparty.

Fixes #1119 

---

### DIFF
--- a/copyparty/web/browser.js
+++ b/copyparty/web/browser.js
@@ -3760,7 +3760,7 @@ function sortfiles(nodes) {
 					if ((v + '').indexOf('<a ') === 0)
 						v = v.split('>')[1];
 					else if (name == "href" && v)
-						v = uricom_dec(v);
+						v = uri2txt(v, true);
 
 					nodes[b]._sv = v
 				}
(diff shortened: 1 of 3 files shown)


copyparty/web/browser.js
```diff
@@ -7873,7 +7873,7 @@ var treectl = (function () {
 		delete res['a'];
 		var keys = Object.keys(res);
 		for (var a = 0; a < keys.length; a++)
-			keys[a] = [uricom_dec(keys[a]), keys[a]];
+			keys[a] = [uri2txt(keys[a]), keys[a]];
```
9001, line 7876: did you mean to set unslash=true, or is that not necessary here?

9001:
neat! just one curiosity and this looks good, and good job spotting why the sorting was off o/

copyparty/web/browser.js
NecRaul: It wasn't necessary. Breadcrumbs menu doesn't append slash to the end.

PR #1530, opened 2026-06-16
nix: use ffmpeg-headless

nyakase:
This PR complies with the DCO; https://developercertificate.org

---

I noticed the default build has a gigantic closure size of ~1.6GiB and wanted to reduce it, since the majority came from graphical libraries.

Swapping `ffmpeg-full` for `ffmpeg-headless` brings the closure size down to ~600MiB. Since late 2024 `ffmpeg-headless` has used `withWebp` and `withOpenmpt` (see https://github.com/NixOS/nixpkgs/commit/193b98084a08ed2b8b55880f77f0525ed915e16e and https://github.com/NixOS/nixpkgs/commit/2fa42d7b13c4b9c39aa19a5680b1ca81faed99a2 respectively) so the reasons in favor of `ffmpeg-full` in the README do not apply anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2636,8 +2636,6 @@ requires a [flake-enabled](https://nixos.wiki/wiki/Flakes) installation of nix
 
 some recommended dependencies are enabled by default; [override the package](https://github.com/9001/copyparty/blob/hovudstraum/contrib/package/nix/copyparty/default.nix#L3-L22) if you want to add/remove some features/deps
 
-`ffmpeg-full` was chosen over `ffmpeg-headless` mainly because we need `withWebp` (and `withOpenmpt` is also nice) and being able to use a cached build felt more important than optimizing for size at the time -- PRs welcome if you disagree 👍
-
 
 ## nixos module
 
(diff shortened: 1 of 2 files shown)


9001:
looks good, just pinging @toast003 and/or @chinponya for a quick opinion since they're the nix gurus (I don't really know anything about this stuff)

also, good time to ask -- now that copyparty is in the main nixos repo (iiuc), can some of the nix stuff in the copyparty repo be deleted, or does it still serve a purpose?

JustTNE:
> also, good time to ask -- now that copyparty is in the main nixos repo (iiuc), can some of the nix stuff in the copyparty repo be deleted, or does it still serve a purpose?

The binary of copyparty is in the nixos repos now, but the "module" - basically what you use to configure copyparty on nixos - is not available there, which is a huge and important part of the nixos stuff in your repo. (I personally rely on the copyparty module for both the package itself as well as the module)

Although it would be possible to remove the package code from your repo, I would suggest keeping it as long as someone is here to maintain it for the simple reason that it is always up to date with the very latest changes to this repo and always in sync with the module version since both are from the same commit.

PR #1535, opened 2026-06-28
Cbz page param and 'go to page'

romfir:
Hi! Thanks for Copyparty :)

I've added 2 small features when reading CBZ files

1. p param to url that updates when changing pages, so its possible to share links to specific page
image

2. 'Go to page' button with 'G' shortcut
image


9001:
thanks! simplified it a bit; think we'll survive without manual changes to the url-bar taking effect :>

PR #1541, opened 2026-07-05
Decode On the Fly!

ptweezy:
Browsers cannot decode every codec/container and I just realized how annoying that little fact is after installing this cool little app. I realized that others, including the maintainer, had this same desire in issue [#224](https://github.com/9001/copyparty/issues/224).

This branch implements the general approach (FFmpeg -> segmented HLS -> hls.js) and differs just a lil' bit from the design sketched in the issue:

- one MPEG-TS segment per index with muxed H.264 video + AAC audio; the issue proposed separate audio and video files for Opus/AAC mix-and-match. Can still enhance this....there's still time!!! :)
- an adaptive-bitrate ladder rather than one server-configured quality: the server generates a rendition per rung (480/720/1080/1440/2160 below the source/`vt_maxh` cap, plus the cap itself) and the browser picks one by screen size + measured bandwidth. The `vt_*` still sets quality ceiling
- segments are transcoded on demand with `-ss` before `-i` (keyframe seek), which matches the issue's snap-to-boundary intent; `-output_ts_offset` places each independent IDR-starting segment on the timeline

## Server

- New module `copyparty/hls.py` (`HlsSrv`): broker service in the hub process (`self.hlssrv`), started only when transcoding is enabled and the required encoders are present.
- Broker verbs: `ensure` (enqueue / return cache dir) and `poke` (keep-alive). `ensure` takes a `height` arg selecting the master playlist (0), a rendition playlist (`height>0`, `idx<0`), or a segment (`height>0`, `idx>=0`).
- Worker pool: `max(1, vt_jobs)` threads pull jobs from a bounded queue (size `vt_jobs*4`), deduped via a busy dict keyed on cache dir + height + segment index; jobs dropped when the queue is full.
- Endpoints:
  - `<source-video>/.hls/master.m3u8`
  - `<source-video>/.hls/<height>/index.m3u8`
  - `<source-video>/.hls/<height>/vNNNNN.ts`
- Shared probe (`_meta`): ffprobe requires a video stream, reads duration, resolution, and HDR flag once, and caches them to `meta.txt` (`resw resh hdr dur`) so the master, every rendition playlist, and each segment reuse it.
- Master playlist (`_gen_master`): computes the rendition ladder (`_ladder`) and writes an ABR master (`#EXTM3U`, `#EXT-X-VERSION:3`, one `#EXT-X-STREAM-INF` per rung with `BANDWIDTH`+`RESOLUTION`, pointing at `<height>/index.m3u8`). Written atomically.
- Rendition playlist (`_gen_playlist`): per-height VOD manifest (`EXT-X-VERSION:3`, `EXT-X-TARGETDURATION`, `MEDIA-SEQUENCE:0`, `PLAYLIST-TYPE:VOD`, `INDEPENDENT-SEGMENTS`, per-segment `#EXTINF`, `EXT-X-ENDLIST`) so the full seekbar is present before segments exist.
- Segment (`_gen_segment`): ffmpeg with `-ss` before `-i` (keyframe seek), `-t <seg>`, `-map 0:v:0 -map 0:a:0?`, scale to even dimensions at the rendition height (`_out_dims`), the encoder's own quality flags (`enc_argv`), `-pix_fmt yuv420p`, `-c:a aac -b:a <vt_aq>k -ac 2`, `-muxdelay 0 -muxpreload 0`, `-output_ts_offset <start>`, `-f mpegts`; output moved into place atomically. ffmpeg is wrapped by bwrap where available.
- `httpcli.py`: `RE_HLS` route (now matching `master.m3u8`, `<height>/index.m3u8`, `<height>/vNNNNN.ts`) and `tx_hls`/`_tx_hls_master`/`_tx_hls_playlist`/`_tx_hls_segment` handlers. `tx_browser` dispatches `.hls` requests when `have_x264 && have_aac && !no_vcode`. The master request polls the cache file up to 30s (breaks as soon as the atomically-written file appears; 504 on timeout, 500 on broker failure); the rendition playlist polls up to 30s for `#EXT-X-ENDLIST`. Child URIs (variant playlists / segments) are rewritten to carry `?k=<filekey>` when present via the shared `_tx_m3u8`. The segment request polls up to 45s, then streams via `tx_file`.
- Permission gate mirrors thumbnails: 404 when `dvcode` is set, extension not in `--th-r-ffv`, stat fails, or path is a directory; filekey evaluated via `_use_filekey`; requires `can_read` or (`can_get` and (filekey or no `fk` flag)) else 403; re-checked on the master, every rendition playlist, and every segment.

### Encoders

- Encoder selection (`_pick_enc` / `enc_argv`): the abstract `vt_enc` value resolves to a real ffmpeg encoder. `auto` uses software `libx264` (best quality-per-bit) and only reaches for a hardware encoder above 1440p, where x264 may not sustain realtime; `x264` forces software; `vt`/`nvenc`/`qsv`/`amf` force the named hardware encoder. Each hardware encoder's own quality/speed knobs are translated from the abstract `(crf, preset)`.
- Boot-time validation (`probe_hwenc` / `ff_test_enc`): at startup each hardware h264 encoder is confirmed to both exist and actually encode one lavfi frame (guards against no-GPU/driver/stub builds); the validated set is cached on the hub as `args.vt_hwenc`.
- Runtime fallback: if a hardware encode fails mid-playback (device busy/driver/unsupported input), the encoder is blacklisted hub-wide (`hw_bad`) and that segment is re-encoded with x264; already-cached hardware segments stay valid.

### HDR

- HDR→SDR tonemapping (`is_hdr` / `tm_vf` / `tm_devargs`): HDR sources (PQ/HLG transfer or BT.2020 primaries) are tonemapped down to SDR BT.709 during transcode so they don't look washed-out on SDR displays.
- Method probing (`probe_tonemap` / `ff_have_filter`): the best method the local ffmpeg can actually run is picked at startup from `libplacebo` (Vulkan), `tonemap_opencl` (OpenCL), and `zscale` (CPU), validated with a synthetic HDR frame; cached on the hub as `args.vt_tm`. If none work, HDR passes through and a warning is logged. `vt_tonemap=off` disables it.

### Caching

- Location: `.hist/vt/<sharded>/<hash>.<mtime-hex>`, a per-source-file directory holding `master.m3u8`, `meta.txt`, and a `<height>/` subdir per rendition (each with its own `index.m3u8` and `vNNNNN.ts`). Keyed on source dir/filename hashes and source mtime. Edits will invalidate.
- Config fingerprint: `.hist/vt/cfg.txt` written from `hls_cfg()` (`vt_maxh`/`vt_vq`/`vt_preset`/`vt_aq`/`vt_seg`/`vt_enc`/`vt_tonemap`); the `th_srv` cleaner wipes the entire `vt` cache when a quality knob changes.
- Expiry: `th_srv._clean` handles category `vt`; per-file dirs matching `PTN_VT` are removed once age exceeds `vt_maxage`.
- Keep-alive: master/playlist/segment requests `utime` the cache dir and up to 4 parent dirs to prevent mid-playback expiry.

## UI/Client

Client-side logic in `baguettebox.js`, gated on the `have_vcode` flag.

- Video detection: when `have_vcode`, `re_v` is widened to include `avi,wmv,flv,ts,m2ts,mts,mpe?g,vob,ogm,rm,rmvb,divx,asf,3gp` in addition to `webm,mkv,mp4,m4v,mov`.
- Native-first: `mp4/mov/webm/m4v` (and `re_vck` on non-Safari) play natively first; transcode is a fallback, so browser's decoder is used when it can be
- Upfront transcode: `re_vhls` containers (`avi,wmv,flv,vob,ogm,rm,rmvb,divx,asf`) are transcoded in auto mode and don't wait for native failure
- `re_vck` containers (`mkv,ts,m2ts,mts,mpe?g,3gp`) may play natively but can silently fail to decode in Chrome. Handled this three ways in auto mode: Safari/iOS with native HLS transcode upfront; `loadedmetadata` with `videoWidth==0` transcodes; a 4000ms `readyState` watchdog transcodes if metadata never loads (`readyState < 1`).
- `load_hls`: points the element at `<source>/.hls/master.m3u8` (the ABR master; the player picks a rendition by screen size + bandwidth). Safari/iOS with native HLS set `src` directly; otherwise `ensure_hls_js` lazy-loads `hls.light.js` and uses `new Hls({ capLevelToPlayerSize: true })` (keeps the auto-picked rendition no larger than the video element) with `MANIFEST_PARSED -> play()`.
- `lerr`: on a native `error` event for a video with `rawsrc` and `!hls_tried`, switches to transcode; guarded against looping when the HLS stream itself fails.
- Source-override toggle (`#bbox-vcode`): cycles auto -> conv -> orig, persisted as `vsrc`. Re-points the current video, tears down any hls instance and watchdog, and restores `currentTime` and play/pause state across the switch; forced `orig` sets `hls_tried=1` to opt out of auto-fallback.
- Teardown: on hide/swap, watchdog timeouts are cleared and we call `hls.destroy()`

## Config

CLI flags (`add_transcoding`):

- `--no-vcode` — disable on-the-fly video transcoding (volflag `dvcode`).
- `--vt-maxh PX` (720) — max transcoded video height; `0` keeps source resolution (volflag `vt_maxh`).
- `--vt-vq CRF` (23) — h264 CRF, lower is better (volflag `vt_vq`).
- `--vt-preset P` (veryfast) — x264 preset; invalid values fall back to veryfast (volflag `vt_preset`).
- `--vt-enc E` (auto) — video encoder: `auto` (software x264 up to 1440p, hardware for 4K), or force `x264`/`vt`/`nvenc`/`qsv`/`amf` (volflag `vt_enc`).
- `--vt-tonemap T` (auto) — tonemap HDR→SDR when transcoding; `auto` (on when ffmpeg has a working tonemap filter) or `off` (volflag `vt_tonemap`).
- `--vt-aq KBPS` (128) — AAC bitrate (volflag `vt_aq`).
- `--vt-seg SEC` (4.0) — target HLS segment duration (volflag `vt_seg`).
- `--vt-maxage SEC` (86400) — delete cached output after SEC seconds.
- `--vt-jobs N` (2) — max concurrent transcodes (`HlsSrv` worker count; not a volflag).

Volflags: `dvcode`, `vt_maxh`, `vt_vq`, `vt_preset`, `vt_enc`, `vt_tonemap`, `vt_aq`, `vt_seg`.

`cfg.py` registers `no_vcode`->`dvcode` in `vf_bmap` and `vt_maxh`/`vt_vq`/`vt_preset`/`vt_enc`/`vt_tonemap`/`vt_aq`/`vt_seg` in `vf_vmap`, and adds thumbnail-section flag help text for `dvcode` and the `vt_*` knobs.

`util.py` adds mime mapping `m3u8=vnd.apple.mpegurl`.

## Reqs

- FFmpeg and FFprobe must be present; otherwise `--no-vcode` is forced at startup.
- FFmpeg must have both `libx264` and `aac` encoders, detected at boot via `ff_have_enc` (`ffmpeg -h encoder=<name>`); if either is missing, transcoding stays disabled and `have_vcode` is false.
- Hardware encoders (`vt`/`nvenc`/`qsv`/`amf`) and HDR tonemapping (libplacebo/opencl/zscale) are optional; each is probed at boot and used only if the local ffmpeg can actually run it, otherwise transcoding falls back to software x264 / SDR passthrough.
- The client `have_vcode` flag (emitted per volume in `authsrv.py`) is `not no_vcode AND have_x264 AND have_aac AND 'dvcode' not in volflags`; setting the `dvcode` volflag on a volume forces `have_vcode` false and disables the client HLS UI there.
- Client requires the bundled `hls.light.js` for non-Safari browsers. Safari uses native HLS.


9001:
no

PR #1577, opened 2026-07-26
Add thumbnailing for KRA and ORA files

Wuerfel21:
(KRA is the extension for Krita's work format, IYDK)

Just grabs the `mergedimage.png` out of the file, very simple. There's also a `preview.png` that's already thumbnail-sized that could be used, but copyparty's hires thumbnails are larger than that and it would break those. And you can just put huge PNGs through the thumbnailer normally, anyways.


Wuerfel21:
Added ORA file extension for this (OpenRaster), since it turns out to have the same mergedimage.png thing going on. Also rebased the branch.

9001:
sweet, thx!

PR #1604, opened 2026-08-09
feat: allow env var for pw in u2c.py

shermanhlc:
for the `u2c.py` script, allows grabbing password from an env var instead of direct from command line or a file. this is helpful for use with systemd and `EnviornmentFile`. not really the _most_ useful, but i wanted it so...

just a simple demo, for sanity's sake
image

### DIFF
--- a/bin/u2c.py
+++ b/bin/u2c.py
@@ -1576,7 +1576,7 @@ NOTE: if server has --usernames enabled, then password is "username:password"
     ap.add_argument("url", type=unicode, help="server url, including destination folder")
     ap.add_argument("files", type=files_decoder, nargs="+", help="files and/or folders to process")
     ap.add_argument("-v", action="store_true", help="verbose")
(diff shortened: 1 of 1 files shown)


9001:
cool idea, will merge it 👍 but regarding the env.var name, to avoid collisions, would `U2C_PW` be better? 

was first thinking of `PRTY_PW` or similar but it also feels like a good idea to semantically disconnect it from anything server-related, since otherwise someone might assume the presence of that env.var would affect copyparty itself

shermanhlc:
> cool idea, will merge it 👍 but regarding the env.var name, to avoid collisions, would `U2C_PW` be better?
> 
> was first thinking of `PRTY_PW` or similar but it also feels like a good idea to semantically disconnect it from anything server-related, since otherwise someone might assume the presence of that env.var would affect copyparty itself

good point. i only did `PASSWD` because that was the metavar and i just assumed thats how it worked when i tried to use it, only to realize that was just a descriptor. 356ef4b

PR #1610, opened 2026-08-18
add jxl support for autogrid

sylfn:
jxl is treated as an image in this regex even if it's unavailable, same as already present webp

i have also removed duplicate svg entries in the regexp

tested on librewolf 153 (like firefox) with jxl support enabled

---


9001:
ah nice thx 👍

PR #1633, opened 2026-09-02
httpcli: reject malformed request headers

Starmania:
Was comparing the output of multiple others origins, and came across some MUST NOT breakage (space and more were allowed in header name, more on header value).

See RFC 9110 ([5.5](https://www.rfc-editor.org/info/rfc9110/#section-5.5)) and 9112 ([5.1](https://www.rfc-editor.org/info/rfc9112/#section-5.1)) for more details

---


9001:
handled by f00d19c5d3ac3004ef318f2927c2adbae3b33f43

PR #1643, opened 2026-09-18
Suppport syntax highlighting based on shebang

### LINKED ISSUE #750: Feature request: Shebang interpretation
**is your feature request related to a problem? Please describe.**
In the browser-based doc viewer, copyparty seems to only use syntax highlighting on files with extensions applied to them; files without extensions on them do not render with any syntax highlighting, regardless if there's a shebang that implies it gets executed with a certain script.

A file (lets say its named simply `main`) like this;
```py
#!/usr/bin/env python3
for i in range(1, 100):
    print(i)
print(":D")
```
...would render with pretty colors instead of white text.

**Additional context**
I am currently running copyparty v1.19.6 within Docker on a Debian computer.

For clarity, here is a screenshot of how the page currently renders.
Image

I'm not totally sure how reasonable this is; I know media files are scanned automatically to propagate their tags within the file browser, and deduplication involves hashing the file serverside, but this may require more involvement (i didn't have a lot of time this afternoon and wasn't able to identify in the source where these things happen). It's obviously very very low priority, but something I noticed that would be a fun addition, so I leave it up to you. :)

botahamec:
Resolves: https://github.com/9001/copyparty/issues/750

This PR allows copyparty to use shebangs to help determine the file type for syntax highlighting.
image

This is my first PR here, so let me know if I missed anything.


9001:
neat, thx! just tweaked it slightly to spot a few more interpreters o/

PR #1647, opened 2026-09-20
fix stale share files when reusing a deleted sharekey

willie6666:
Deleting a share through handle_eshare() currently removes its row from sh but leaves matching rows in sf.

If the same sharekey is later reused for a file share, the stale sf rows are loaded together with the new files, causing files from the deleted share to reappear.

Expired-share cleanup already removes rows from both sh and sf, so this change makes manual deletion behave consistently.


9001:
ah nice catch, thx!

